PR #1, opened 2020-03-15
Admin panel

tyxchen:
Buncha admin panel functionality, i.e. managing users


tyxchen:
We're going to have to figure out how to reconcile the localStorage auth solution with the promise-based auth solution. We could adopt one, or both. Thoughts?

src/App.tsx
```diff
@@ -1,48 +1,149 @@
...
+          </Switch>
         </div>
-      </BrowserRouter>
-    </AuthProvider>
+      </div>
+    </Router>
+  );
+};
+
+const ToastContainer = () => {
```
ktrieu: Should move this into the shared/ directory.

src/App.tsx
```diff
@@ -1,48 +1,149 @@
...
+  const toast = useToast();
+
+  return (
+    <div aria-live="polite" aria-atomic="true" style={{
+      position: "fixed",
+      right: 0,
+      bottom: 0,
+    }}>
+      {toast.toasts.map((msg, i) => {
```
ktrieu: I respect this algorithm you've built here, but instead of doing what I believe is a linear search, you could just use a map.

src/profile/ProfileForm.tsx
```diff
@@ -0,0 +1,494 @@
...
+  { type: 'is checking' } |
+  { type: 'done checking' } |
+  { type: 'done loading success' } |
+  { type: 'done loading error', errors: UserAPIError | string | string[] } |
+  { type: 'set error', errors: UserAPIError } |
+  { type: 'set general error', errors: string[] } |
+  { type: 'set success message', message: string } |
+  { type: 'set data', data: ChangedUser, errors?: UserAPIError };
+
+const profileReducer = (state: ProfileState, action: ProfileAction): ProfileState => {
```
ktrieu: I anticipate us having a lot of forms, we should use Formik instead of handrolling a reducer for every single one.

src/shared/errors.ts
```diff
@@ -0,0 +1,78 @@
+const errorFactory = (templ: (...prop: string[]) => string) => new Proxy({}, {
```
ktrieu: This is another thing we could replace with Formik.

src/shared/errors.ts
```diff
@@ -0,0 +1,78 @@
...
+  get(_, prop: string) {
+    return templ(...prop.replace('_', ' ').split(','));
+  },
+  has(_, prop) {
+    return true;
+  }
+});
+
+const errors = {
```
ktrieu: Maybe just bite the bullet and use an i18n library for this, it's not that much extra work, and we get interpolation in translated strings for free, pretty much.

src/shared/types.ts
```diff
@@ -25,9 +25,40 @@ export interface User {
...
+  detail?: string[];
+  user?: string[];
+  username?: string[];
+  email?: string[];
+  password?: string[];
+  writer_name?: string[];
+  [key: string]: string[] | undefined;
+}
+
+export interface UserAPIResponse {
```
ktrieu: Like the idea, but we should make it generic and use it API-wide. I.E, `APIResponse<T> { error: <some error obj>, data: T, success: boolean }`

src/shared/errors.ts
tyxchen: I've taken a look at the i18n libraries around, and to be honest the overall benefit seems marginal -- what we gain in convenience, we lose in package size. Can be something to experiment with in the future though, I think.

src/App.tsx
tyxchen: Because React state depends on comparing objects, using a map wouldn't work, as React would detect the updated state as being the same and wouldn't rerender. To get around this, the DS used needs to be copied, which is an O(n) operation regardless.

I should probably abstract away the heuristics into ToastContext though, it's pretty ugly.

src/App.tsx
tyxchen: 027b8d59

src/App.tsx
tyxchen: 0e0707b3

src/shared/types.ts
tyxchen: 57443008

In the future, we probably should change the API to return the user in the data field.

ktrieu: DRF let's you customize the final output of all API endpoints. I did the same trick at a co-op once, I'll look into it.

ktrieu:
Think we're good here. Do you want to to handle the merge conflict or do you want to me to take it?

tyxchen:
I think we're good too. Trying to get Formik to work, but I'm having trouble implementing debouncing for the fields that require server-side validation; I think I'll take a closer look tomorrow at my options.

The merge conflicts are pretty complicated and extend beyond the files listed. I could do them, but it'd have to be tomorrow, so if you want this merged soon you should handle it. Otherwise I can take a shot.

PR #3, opened 2020-03-17
Style revamp

ktrieu:
Change the styling and layout to conform to our mockups.


src/editor/EditorControls.scss
```diff
@@ -1,6 +1,5 @@
 .editor-controls {
   padding: 5px;
```
tyxchen: Thought: We should probably have some sort of standardized unit size and set that as a variable, then scale that unit accordingly. Right now the mockups are based on 4px spacing, so that might be a good place to start.

src/editor/components/controls.scss
```diff
@@ -13,6 +13,10 @@
     height: 100%;
     max-width: 100%;
     width: auto;
+
+    & > path {
+      fill: #4f4f4f;
```
tyxchen: This should be a variable of some sort.

src/header/SluglineNav.scss
```diff
@@ -1,14 +1,22 @@
...
+    font-weight: normal;
+  }
+}
 
-nav.blackbox {
-  background: black !important;
-  padding: 0.5em 0.5em;
+#navContainer {
+  padding-left: 15px;
+  padding-right: 15px;
```
tyxchen: See comment on units.

src/header/SluglineNav.tsx
```diff
@@ -2,47 +2,61 @@ import React from "react";
...
+}
+
+// React Router's links and Bootstrap's navbar links don't play together too nicely,
+// so try and mash them together here
+const NavLink: React.FC<NavLinkProps> = props => {
+  const { text, ...rest } = props;
+  return (
+    <Nav.Item>
+      <Link {...rest} className="nav-link">
+        <h4>{props.text}</h4>
```
tyxchen: For semantic reasons, we should probably style the links in CSS rather than using a heading element.

src/editor/EditorControls.scss
ktrieu: I *believe* Bootstrap has some kind of default spacing variable. We can set that to 4px, and then go from there.

ktrieu: Well turns out, $spacer = 1rem = 16px, and the first spacing increment is 0.25 * $spacer = 4px. Sometimes life does work out that way.

tyxchen: Found it: https://getbootstrap.com/docs/4.4/getting-started/theming/#sass-options

Looks like the default spacing variable (`$spacer`) is meant for margins and so is set to a relatively large value (1rem). We may just want to set it to 16px so that the minimum spacing unit is 4px; alternatively, we can set `$spacer` to 4px, and change the size mappings in the `$spacers` map (https://getbootstrap.com/docs/4.4/utilities/spacing/#notation).

ktrieu: Well, `mb-1` is basically 4px, since its `$spacer * 0.25`. Think we can leave the mapping the way it is for now.

src/editor/EditorControls.scss
```diff
@@ -1,11 +1,13 @@
+@import "src/slugline.scss";
+
 .editor-controls {
-  padding: 5px;
+  padding: map-get($map: $spacers, $key: 1);
```
tyxchen: Thought: in the future, might be worth defining a helper function for getting spacing.

ktrieu: I'll get that in before we merge.

PR #6, opened 2020-04-27
Add support for article editing flow

ktrieu:
This PR adds two dashboard pages: a list of the user's articles and a page that displays the editor. It also adds support for automatic saving of article edits.

### DIFF
--- a/src/api/hooks.ts
+++ b/src/api/hooks.ts
@@ -13,7 +13,7 @@ import {
   APIResponse,
   ArticleContent,
 } from "../shared/types";
-import { useState, useEffect } from "react";
+import { useState, useEffect, useCallback } from "react";
 import { useAuth } from "../auth/AuthProvider";
 import { getApiUrl } from "./api";
 
@@ -67,23 +67,25 @@ const useApiPost = <S, T, U extends APIError = APIError>(
 
   const [state, setState] = useState<RequestState>(RequestState.NotStarted);
 
-  let headers: { [header: string]: string } = {};
-  if (auth.csrfToken) {
-    headers["X-CSRFToken"] = auth.csrfToken;
-  }
-
-  const post = async (body: S): Promise<T> => {
-    setState(RequestState.Started);
-    const resp = await axios.post<APIResponse<T>>(url, body, {
-      headers: headers,
-    });
-    setState(RequestState.Complete);
-    if (resp.data.success) {
-      return resp.data.data;
-    } else {
-      throw resp.data;
-    }
-  };
+  const post = useCallback(
+    async (body: S): Promise<T> => {
+      let headers: { [header: string]: string } = {};
+      if (auth.csrfToken) {
+        headers["X-CSRFToken"] = auth.csrfToken;
+      }
+      setState(RequestState.Started);
+      const resp = await axios.post<APIResponse<T>>(url, body, {
+        headers: headers,
+      });
+      setState(RequestState.Complete);
+      if (resp.data.success) {
+        return resp.data.data;
+      } else {
+        throw resp.data;
+      }
+    },
+    [url, auth.csrfToken]
+  );
 
   return [post, state];
 };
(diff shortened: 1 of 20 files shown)


ktrieu:
@tyxchen There are some leftover warnings to do with missing `useEffect` dependencies in the `auth` stuff. Don't think I have the knowledge to evaluate that, so could you take a look and fix those up?

src/api/hooks.ts
```diff
@@ -0,0 +1,159 @@
...
+  Article,
+  Pagination,
+  APIResponse,
+  ArticleContent,
+} from "../shared/types";
+import { useState, useEffect } from "react";
+import { useAuth } from "../auth/AuthProvider";
+import { getApiUrl } from "./api";
+
+const useApiGet = <T, U extends APIError = APIError>(
```
tyxchen, line 20: At some point, we'll need to add documentation. Might be a good idea to start that earlier than later.

src/api/hooks.ts
```diff
@@ -0,0 +1,159 @@
...
+};
+
+const useApiPost = <S, T, U extends APIError = APIError>(
+  url: string
+): APIPostHook<S, T, U> => {
+  const auth = useAuth();
+
+  const [state, setState] = useState<RequestState>(RequestState.NotStarted);
+
+  let headers: { [header: string]: string } = {};
```
tyxchen, line 70: Can we make this a general type? I see this being a pattern we may encounter a lot in the future.

src/shared/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (timeout.current) {
+      clearTimeout(timeout.current);
+    }
+  }
+
+  // make sure our timeout gets cleared if
+  // our consuming component gets unmounted
+  useEffect(() => cleanup, []);
+
+  return function debouncedCallback(...args: A) {
```
tyxchen, line 21: Any reason for a named function instead of an anonymous one here? Just curious.

src/shared/hooks.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useRef } from "react";
+
+export function useDebouncedCallback<A extends any[]>(
```
tyxchen, line 3: Should this be `export const ...` for consistency?

src/shared/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+  callback: (...args: A) => void | Promise<void>,
+  wait: number
+) {
+  // track args & timeout handle between calls
+  const argsRef = useRef<A>();
+  const timeout = useRef<ReturnType<typeof setTimeout>>();
+
+  function cleanup() {
```
tyxchen, line 11: Ditto with const function

tyxchen:
Sidenote: Can you commit your editor/formatter settings to the repo? That way we won't have to deal with it in the future.

src/shared/hooks.ts
ktrieu: Full disclosure: I copied and pasted this from StackOverflow.

tyxchen:
Ok, might want to link to the answer from which the code was taken.

src/shared/hooks.ts
tyxchen: Sorry, I replied to this via email but it turns out GitHub just creates a new comment instead.

It's advisable to link the the SO answer that you got this from, then. They license snippets under CC-BY SA 4.0 so attribution is required.

ktrieu:
I haven't touched any of the default settings, I think you just need to install the Prettier extension for VS Code. (and turn on format on save)

src/api/hooks.ts
ktrieu: So, I put some more thought into this, and since there are really only four-ish HTTP methods we care about, and and only two of them are the same really, I don't think we should bother trying to abstract this.

tyxchen:
> @tyxchen There are some leftover warnings to do with missing `useEffect` dependencies in the auth stuff. Don't think I have the knowledge to evaluate that, so could you take a look and fix those up?

We can safely ignore those warnings. The reason why the `exhaustive-deps` warning exists is to avoid cases where the effect doesn't run on state change. We want the opposite of that --- our auth effects should only run on app mount --- so we provide an empty dependency array, but that causes eslint to complain. 

Something I discovered when investigating this though: because `AuthProvider` stores a promise in state, which is changed whenever `auth.check` is called, `useAuth()` unwittingly changes every time we call `auth.check`. This probably will have performance/concurrency implications later on, so I'll file an issue about it.

PR #8, opened 2020-05-05
Add an list of issues to the dashboard for admins

ktrieu:
This PR adds an list of issues to the dashboard for admins, as well as a button to create new ones.

The issue creation form uses `react-hook-form`, as well as some common form code introduced in this PR that I hope will be used in future. 

This PR also includes some refactoring of the API hooks to return `APIResponses` so that calling code can handle errors without exceptions, and some refactoring of the SCSS to prevent CSS rules being included multiple times.


src/_slugline_vars.scss
```diff
@@ -13,3 +13,20 @@ $link-color: black;
 // Navbar variables
 $navbar-light-color: $primary;
 $navbar-brand-font-size: 2rem;
+
+// Forms
+$input-border-radius: 0;
+$input-border-width: 1px;
+$input-border-color: black;
+$input-focus-box-shadow: none;
+$input-focus-border-color: $primary;
```
tyxchen, line 22: We should look into reducing the horizontal padding in text inputs here.

src/_slugline_vars.scss
```diff
@@ -13,3 +13,20 @@ $link-color: black;
 // Navbar variables
 $navbar-light-color: $primary;
 $navbar-brand-font-size: 2rem;
+
+// Forms
+$input-border-radius: 0;
+$input-border-width: 1px;
+$input-border-color: black;
```
tyxchen, line 20: Should be a variable.

(sidenote: I think black is too black, would prefer the original bootstrap colour)

src/dash/Dash.tsx
```diff
@@ -11,7 +13,13 @@ const Dash = () => {
       <Route path={`${match.path}/edit/:articleId`}>
         <EditorPage />
       </Route>
-      <Route path={`${match.path}/`}>
+      <Route path={`${match.path}/issues/:issueId`}>
+        <DashIssueDetail />
+      </Route>
+      <Route path={`${match.path}/issues/`}>
+        <DashIssuesPage />
```
tyxchen, line 20: With the introduction of these routes, we need to (1) remove the routes in App.tsx, and (2) change the links in the header.

src/dash/articles/DashArticlesPage.tsx
```diff
@@ -1,17 +1,17 @@
 import React from "react";
-import { useUserArticles, useCreateArticle } from "../api/hooks";
+import { useUserArticles, useCreateArticle } from "../../api/hooks";
```
tyxchen, line 2: thought: should probably configure webpack imports to add a prefix for local imports (so instead of `../../api/hooks`, we could just write `src/api/hooks`)

src/dash/issues/DashIssueDetail.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { ArticleTitle } from "../articles/DashArticlesPage";
+
+const DashIssueDetail = () => {
+  const { issueId } = useParams();
+
+  const id = parseInt(issueId || "");
+
+  const [issue] = useIssue(id);
+  const [resp] = useIssueArticles(id);
```
tyxchen, line 14: At some point we're going need to handle potential errors returned by these hooks.

src/shared/types.ts
```diff
@@ -17,15 +17,15 @@ export interface Article {
...
+  content_raw: string;
 }
 
 export interface Issue {
-  id: number;
-  publish_date: string;
+  id?: number;
+  publish_date?: string;
   volume_num: number;
   issue_num: number;
```
tyxchen, line 27: issue_num should be a string (e.g. for election issues, we would publish `v111iE`).

src/dash/issues/DashIssuesPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+          />
+          <Form.Label>i</Form.Label>
+          <Field
+            name="issueNum"
+            className="issue-num-input"
+            type="text"
+            maxLength={1}
+            ref={register({
+              required: true,
+              pattern: /[0-9]{1}/,
```
tyxchen, line 92: Should allow for an `E` issue (or, in general, any issue consisting of a single capital letter, which I think will cover all special issue cases).

src/dash/issues/DashIssuesPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+            })}
+            errors={errors}
+            hideErrorMessage
+          />
+        </Form>
+        {nonFieldErrors?.map((error, idx) => (
+          <small key={idx} className="invalid-feedback d-block">
+            {ERRORS[error]}
+          </small>
+        ))}
```
tyxchen, line 104: Can we componentize this map? I like it.

src/dash/issues/DashIssuesPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+        <Button variant="secondary" onClick={closeModal}>
+          Close
+        </Button>
+      </Modal.Footer>
+    </Modal>
+  );
+};
+
+const DashIssuesPage = () => {
+  const [resp] = useIssueList();
```
tyxchen, line 119: ditto with having to handle possible errors at some point.

src/profile/ProfileForm.tsx
```diff
@@ -386,13 +386,11 @@ const ProfileForm: React.FC<{
...
-                  {state.errors.username?.map((msg) => (
-                    <li key={msg}>{ERRORS[msg]}</li>
-                  ))}
-                </ul>
-              </Form.Control.Feedback>
+              <ul>
+                {state.errors.username?.map((msg) => (
+                  <li key={msg}>{ERRORS[msg]}</li>
+                ))}
+              </ul>
```
tyxchen, line 393: See remark about componentizing error messages.

src/dash/issues/DashIssuesPage.tsx
tyxchen: (sidenote: if we do componentize this, it would probably be better to use the error as the key instead of the index, since errors should not be duplicated)

src/slugline.scss
```diff
@@ -39,3 +31,14 @@ h5,
 h6 {
   text-transform: uppercase;
 }
+
+// Form styling
+input.form-control {
+  border-top: none;
+  border-left: none;
+  border-right: none;
```
tyxchen, line 39: This breaks input groups:
image

src/dash/issues/DashIssuesPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+            name="volumeNum"
+            className="volume-num-input"
+            type="text"
+            maxLength={3}
+            ref={register({
+              required: true,
+              pattern: /[0-9]{3}/,
+              min: 0,
+              maxLength: 3,
+            })}
```
tyxchen, line 80: It would be simpler to make this field `type="number"` and set min/max attributes instead.

src/slugline.scss
```diff
@@ -39,3 +31,14 @@ h5,
...
+
+  &.is-invalid {
+    box-shadow: none !important;
```
tyxchen, line 42: Rule should also be extended to `.is-valid`.

src/dash/Dash.tsx
ktrieu: I think your mockups have a separate list of routes for dashboard and main site stuff, which I think would be better, since most visitors won't ever see the dashboard.

src/dash/issues/DashIssueDetail.tsx
ktrieu: Think we should refactor the get hooks to return the status code somehow. The primary error here is going to be some variation on a 404.

src/dash/issues/DashIssuesPage.tsx
ktrieu: A `<NonFieldError>` or `<SubmissionError>` component makes sense.

src/profile/ProfileForm.tsx
ktrieu: Think we can leave this for now and switch all the forms to the form library in a separate PR.

src/slugline.scss
ktrieu: Yeah, not sure how to fix that. Maybe just make it a separate button.

src/dash/issues/DashIssuesPage.tsx
ktrieu: The maxLength prevents you from typing in more than three numbers, but that doesn't work with `type="number"`, as far as I know.

src/dash/Dash.tsx
tyxchen: Ahh, I see what you're getting at. My bad.

src/dash/issues/DashIssueDetail.tsx
tyxchen: We can put status code in APIError.

src/dash/issues/DashIssuesPage.tsx
tyxchen: `maximum="999"` should work, I believe?

src/slugline.scss
tyxchen: I believe the input group is encased in an `.input-group` class or something like that. We could probably just remove all but the bottom border.

src/dash/issues/DashIssuesPage.tsx
ktrieu: The min/max properties only affect the spinner buttons that increment or decrement the value, they don't actually affect the text input at all.

src/slugline.scss
ktrieu: image

Tried it, and it honestly looks like a floating eye. Think we need some kind of differentiation on the left, but with the left and bottom border it looks weird too.

src/dash/issues/DashIssuesPage.tsx
tyxchen: Huh, TIL. In that case the only thing I can think of is changing the pattern so that leading zeroes aren't allowed.

src/slugline.scss
tyxchen: Came up with this -- thoughts?
image

src/dash/issues/DashIssuesPage.tsx
ktrieu: I think I wasn't clear about the behaviour I wanted, I want the user to be able to only type in three characters at which point the text input will stop accepting keystrokes, and the pattern doesn't affect the actual input, only the browser validation afterwards. AFAIK, the only way to do this is the `maxLength` attribute.

src/slugline.scss
ktrieu: Looks better, but I still think we need some way to distinguish the eye as a button and not just an icon. Maybe another background color?

I do like the background color on the input though, it matches the "subscribe to delivery" section we have at the back of some issues, which is the only reference we have for form inputs.

src/_slugline_vars.scss
ktrieu: Cut the padding in half, let me know what you think about it.

src/dash/articles/DashArticlesPage.tsx
ktrieu: This would probably require messing with `create-react-app` which doesn't sound too appetizing. I'm personally OK with leaving them like this.

src/slugline.scss
tyxchen: Yeah, the disabled aspect is not appealing...I'm thinking we can leave it as is right now, honestly I'm fine with the floating eye look --- we can also have the bottom border match colours in CSS later on.

Alternatively, we can keep the default Bootstrap styling.

src/dash/issues/DashIssuesPage.tsx
tyxchen: You're right on the `maxLength`. Just to prevent headaches later, I'd still suggest modifying the validation pattern to disallow leading zeros though.

This is for later, but it would be nice if focus could be transitioned to the issue text box once the third digit is entered (or `i` is pressed).

src/slugline.scss
ktrieu: So I just consulted Material UI, (which is what we're trying to copy here anyyway), and...
they just went with the floating eye. Let's go with it.

PR #9, opened 2020-05-07
Add basic CI

vincentmacri:


.github/workflows/master.yml
```diff
@@ -0,0 +1,33 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
```
tyxchen, line 18: Can this be changed to Debian Stretch? It's what we currently run on.

vincentmacri: GitHub actions do not support Debian Stretch.

I can change it to Ubuntu 16.04, which is based off of Stretch according to https://askubuntu.com/questions/445487/what-debian-version-are-the-different-ubuntu-versions-based-on

tyxchen: Yeah, that sounds good. Took a cursory look and most package versions are the same as on debian, which is good.

vincentmacri:
This fixes #11

vincentmacri:
@tyxchen approval to merge?

PR #10, opened 2020-05-07
Add a component for displaying tabular data

tyxchen:
This PR introduces the `useRichTable` hook and the `RichTable` component, which are designed to display tabular data retrieved from an API endpoint. Thus, all interaction with the API can be done through `useRichTable`, allowing for extremely clean implementation without needing to keep track of mundane details such as pagination and searching. RichTable, being a wrapper around the hook, offers the same flexibility with the added benefit of a ready-to-use table. 

The following are supported:

- Paginated data
- Sorting (server-side)
- Searching (server-side)
- Selecting
- Bulk operations on data

Future iterations may explore a simple interface to filter data, and the possibility of having something like Notion's Views, i.e. saving certain user-specified configurations such as filters and sorting order.


src/shared/components/RichTable.tsx
```diff
@@ -0,0 +1,666 @@
...
+ */
+
+/**
+ * First, we must introduce some basic types.
+ */
+
+export type Accessor<D extends object> = (row: D) => any;
+
+export interface ColumnProps<D extends object = {}> {
+  Header: string;
```
ktrieu, line 41: Any particular reason this is uppercase?

src/shared/components/RichTable.tsx
```diff
@@ -0,0 +1,666 @@
...
+            );
+          }
+        }
+      } else {
+        return action.call(bag);
+      }
+    },
+    // We can ignore the warning about bag being a missing dependency -- although this technically depends on
+    // it, both depend on data and so will refresh appropriately.
+    [data, memoizedActions]
```
ktrieu, line 276: To be frank, I'm not quite comfortable with going against the linter on this. While I think we should strive to get rid of all these cases, at the very least, we should drop an explicit `// eslint-disable-next-line react-hooks/exhaustive-deps` so we don't get used to having a bunch of warnings in the dev server, and if anything breaks, we can go back and look there first.

ktrieu: There's a big list of common "linter complaining at me scenarios" [here](https://github.com/facebook/react/issues/14920#issuecomment-471070149), if that helps any.

src/shared/components/RichTable.tsx
```diff
@@ -0,0 +1,666 @@
...
+    rows,
+    selected,
+    page,
+    setPage,
+    numPages,
+    totalCount,
+    setSearchQuery,
+    executeAction,
+  } = useRichTable(config);
+  const debouncer = useRef<number>(-1);
```
ktrieu, line 538: We do have a debouncer hook in the shared folder. Any reason why that isn't being used here?

src/admin/UserList.tsx
```diff
@@ -45,6 +47,39 @@ const UserList = () => {
   const editUserRef = useRef({ submit: () => Promise.resolve() });
   const createUserRef = useRef({ submit: () => Promise.resolve() });
 
+  const columns = useMemo<Column<User>[]>(
```
ktrieu, line 50: Instead of memo'ing this, we can just pull it out into the top-level, above the component.

src/shared/components/RichTable.tsx
tyxchen: It's a holdover from react-table, which follows the same convention. I was halfway through testing that out when I decided to roll this, hence the capitalization.

src/shared/components/RichTable.tsx
tyxchen: I should note that because of how JS does block-level variable resolution, I can't put `bag` in the dependencies. My worry about ignoring exhaustive-deps here was that it might cover an actual problem later on; I'd rather have a single warning in the dev server for which we have documentation that we can ignore, rather than missing out on legitimate warnings later.

src/shared/components/RichTable.tsx
tyxchen: The behaviour I was going for was an instant refresh when you clear the search query, which isn't possible with the debouncer hook.

src/admin/UserList.tsx
tyxchen: Fair cop. 8f2d1c4e

src/shared/components/RichTable.tsx
ktrieu: We can probably modify it to add that functionality. Something like:
```
export const useDebouncedCallback = <A extends any[]>(
  callback: (...args: A) => void | Promise<void>,
  wait: number
) => {
  // track args & timeout handle between calls
  const argsRef = useRef<A>();
  const timeout = useRef<ReturnType<typeof setTimeout>>();

  const cleanup = () => {
    if (timeout.current) {
      clearTimeout(timeout.current);
    }
  };

  // make sure our timeout gets cleared if
  // our consuming component gets unmounted
  useEffect(() => cleanup, []);

  return [
    useCallback(
      (...args: A) => {
        // capture latest args
        argsRef.current = args;

        // clear debounce timer
        cleanup();

        // start waiting again
        timeout.current = setTimeout(() => {
          if (argsRef.current) {
            callback(...argsRef.current);
          }
        }, wait);
      },
      [callback, wait]
    ),
    // add function to immediately trigger the callback
    useCallback(
      (...args: A) => {
        callback(...args);
        cleanup();
      },
      [callback]
    ),
  ];
};
```

Having a named hook instead of a bare ref should be more clear.

src/shared/components/RichTable.tsx
ktrieu: On the other hand, if warnings pop up in other parts of the code, they might get lost in the middle of the "warnings we're OK with". Not to mention we'll have to get CI to ignore warnings as well.

src/shared/components/RichTable.tsx
tyxchen: 7e0f3c1

src/shared/components/RichTable.tsx
ktrieu: Can it be renamed? Usually we capitalize things because they're going to be JSX, which is misleading especially for a header, which could be either JSX or a string.

ktrieu:
Maybe I'm being an idiot, but I can't get to the admin page on the front end, it keeps pinging me back to a login screen. I'm trying to login as the superuser.

src/shared/components/RichTable.tsx
tyxchen: We can go on an eslint-hunting run later on; that will cut down on the number of warnings that show up, especially since a lot of them have to be silenced. Once we get to that point, we'll be left with only that warning, hopefully.

ktrieu: Fair enough, we've probably spent enough time bikeshedding on this as it is. By the power vested in me by the Province of Ontario, I hereby resolve this conversation.

src/shared/components/RichTable.tsx
tyxchen: Done: 70010b3

PR #13, opened 2020-05-10
Migrate the ProfileForm to react-hook-form

ktrieu:
This PR migrates the ProfileForm to react-hook-form and hopefully sets some conventions for future forms. 

The form looks entirely the same, except for my one innovation: splitting name into two fields.

---

Blocked by: none

Blocks: #21

### DIFF
--- a/src/admin/UserList.tsx
+++ b/src/admin/UserList.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useReducer, useRef, useState } from "react";
+import React, { useRef, useState, useEffect } from "react";
 import { getApiUrl } from "../api/api";
 import { Modal, Button } from "react-bootstrap";
 
--- a/src/admin/UserList.tsx
+++ b/src/admin/UserList.tsx
@@ -79,7 +79,6 @@ const UserEditModal: React.FC<UserEditModalProps> = (
           user={props.user}
           onSubmitSuccessful={() => {
             props.refreshTable();
-            props.setShow(false);
           }}
           hideSubmit
         />
--- a/src/admin/UserList.tsx
+++ b/src/admin/UserList.tsx
@@ -14,6 +14,38 @@ import {
   RichTableBag,
 } from "../shared/components/RichTable";
 
+const columns: Column<User>[] = [
+  {
+    header: "Username",
+    key: "username",
+    sortable: true,
+    width: 15,
+  },
+  {
+    header: "Name",
+    key: "name",
+    width: 20,
+    accessor: (user: User) =>
+      `${user.first_name}${user.last_name ? ` ${user.last_name}` : ""}`,
+  },
+  {
+    header: "Writer Name",
+    key: "writer_name",
+    width: 20,
+  },
+  {
+    header: "Email",
+    key: "email",
+    width: 25,
+  },
+  {
+    header: "Role",
+    key: "role",
+    accessor: (user: User) =>
+      user.is_staff ? "Staff" : user.is_editor ? "Editor" : "Contributor",
+  },
+];
+
 interface UserEditModalProps {
   user?: User;
   show: boolean;
@@ -72,38 +104,6 @@ const UserEditModal: React.FC<UserEditModalProps> = (
     </Modal>
   );
 };
-
-const columns: Column<User>[] = [
-  {
-    header: "Username",
-    key: "username",
-    sortable: true,
-    width: 15,
-  },
-  {
-    header: "Name",
-    key: "name",
-    width: 20,
-    accessor: (user: User) =>
-      `${user.first_name}${user.last_name ? ` ${user.last_name}` : ""}`,
-  },
-  {
-    header: "Writer Name",
-    key: "writer_name",
-    width: 20,
-  },
-  {
-    header: "Email",
-    key: "email",
-    width: 25,
-  },
-  {
-    header: "Role",
-    key: "role",
-    accessor: (user: User) =>
-      user.is_staff ? "Staff" : user.is_editor ? "Editor" : "Contributor",
-  },
-];
 
 const UserList = () => {
   const auth = useAuth();
--- a/src/admin/UserList.tsx
+++ b/src/admin/UserList.tsx
@@ -3,25 +3,75 @@ import { getApiUrl } from "../api/api";
 import { Modal, Button } from "react-bootstrap";
 
 import { User } from "../shared/types";
-import ProfileForm from "../profile/ProfileForm";
+import ProfileForm, { useProfileForm } from "../profile/ProfileForm";
 
 import "./UserList.scss";
 import { useAuth } from "../auth/AuthProvider";
 import ERRORS from "../shared/errors";
-import { RichTable, Column, RichTableBag } from "../shared/components/RichTable";
+import {
+  RichTable,
+  Column,
+  RichTableBag,
+} from "../shared/components/RichTable";
 
-interface State {
-  allUsers: User[];
-  filteredUsers: User[] | undefined;
-  currentUser: User | undefined;
-  showEditUser: boolean;
-  showCreateUser: boolean;
+interface UserEditModalProps {
+  user?: User;
+  show: boolean;
+  setShow: (show: boolean) => void;
+  refreshTable: () => void;
 }
 
-type Action =
-  { type: 'set edit user', data: User | undefined } |
-  { type: 'show edit user', data: boolean } |
-  { type: 'show create user', data: boolean };
+const UserEditModal: React.FC<UserEditModalProps> = (
+  props: UserEditModalProps
+) => {
+  const context = useProfileForm(props.user);
+
+  // grab a reference to the form state here, or else it won't update properly
+  const formState = context.formState;
+
+  return (
+    <Modal
+      show={props.show}
+      onHide={() => {
+        props.setShow(false);
+      }}
+      size="lg"
+    >
+      <Modal.Header>
+        <Modal.Title>{props.user ? "Edit User" : "Create User"}</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <ProfileForm
+          context={context}
+          formId="editUserForm"
+          user={props.user}
+          onSubmit={async () => {
+            props.refreshTable();
+            props.setShow(false);
+          }}
+          hideSubmit
+        />
+      </Modal.Body>
+      <Modal.Footer>
+        <Button
+          variant="secondary"
+          onClick={() => {
+            props.setShow(false);
+          }}
+        >
+          Close without saving
+        </Button>
+        <Button
+          type="submit"
+          disabled={formState.isSubmitting}
+          form="editUserForm"
+        >
+          {formState.isSubmitting ? "Saving..." : "Save"}
+        </Button>
+      </Modal.Footer>
+    </Modal>
+  );
+};
 
 const columns: Column<User>[] = [
   {
@@ -34,7 +84,8 @@ const columns: Column<User>[] = [
     header: "Name",
     key: "name",
     width: 20,
-    accessor: (user: User) => `${user.first_name}${user.last_name ? ` ${user.last_name}` : ""}`
+    accessor: (user: User) =>
+      `${user.first_name}${user.last_name ? ` ${user.last_name}` : ""}`,
   },
   {
     header: "Writer Name",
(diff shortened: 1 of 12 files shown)


src/admin/UserList.tsx
```diff
@@ -1,27 +1,77 @@
...
+  show: boolean;
+  setShow: (show: boolean) => void;
+  refreshTable: () => void;
 }
 
-type Action =
-  { type: 'set edit user', data: User | undefined } |
-  { type: 'show edit user', data: boolean } |
-  { type: 'show create user', data: boolean };
+const UserEditModal: React.FC<UserEditModalProps> = (
```
tyxchen, line 24: nit: Can this be moved after columns?

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-          {state.successMessage}
-        </Alert>
-      )}
-
-      <Form.Group as={Row} controlId="profileUsername">
+    <Form id={props.formId} onSubmit={props.context.handleSubmit(submit)}>
+      <Form.Group as={Row} controlId="username">
         <Form.Label column sm={2}>
-          Username
+          Username:
```
tyxchen, line 118: Colons shouldn't be included in form labels.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+            errors={props.context.errors}
+            type="text"
+            name="last_name"
+            placeholder="Last Name"
+            ref={props.context.register({
+              maxLength: {
+                value: 150,
+                message: "USER.LAST_NAME.TOO_LONG.150",
+              },
+              required: true,
```
tyxchen, line 180: Shouldn't be required; not all people have last names.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-                : "")
-            }
-            onChange={onChange}
-            isValid={"first_name" in state.changedUser}
+          <Field
+            errors={props.context.errors}
+            type="email"
+            name="email"
+            ref={props.context.register({
+              required: "USER.REQUIRED.EMAIL",
```
tyxchen, line 195: `EMAIL` should be lowercased.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+            type="text"
             name="writer_name"
-            placeholder="e.g. Grower of Apples"
-            value={state.changedUser.writer_name ?? user?.writer_name ?? ""}
-            onChange={onChange}
-            isValid={(state.changedUser.writer_name?.length ?? 0) > 0}
-            isInvalid={
-              state.errors.writer_name && state.errors.writer_name.length > 0
-            }
+            ref={props.context.register}
```
tyxchen, line 217: Writer name should be required.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-                <li key={msg}>{ERRORS[msg]}</li>
-              ))}
-            </ul>
-          </Form.Control.Feedback>
         </Col>
       </Form.Group>
-
-      {auth.isEditor() && (
-        <Form.Group as={Row} controlId="profileIsEditor">
+      {auth.user?.is_editor && (
```
tyxchen, line 221: Should use `auth.isEditor()`

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+        <Form.Group as={Row} controlId="isEditor">
           <Form.Label column sm={2}>
-            Role
+            Status:
```
tyxchen, line 224: I think Role is better than Status here --- it sounds more natural, and is consistent with Wordpress conventions + the backend.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+                  props.context.setValue("is_editor", true);
+                } else if (e.currentTarget.value === "contributor") {
+                  props.context.setValue("is_editor", false);
+                }
+              }}
+              value={
+                props.context.getValues().is_editor ? "editor" : "contributor"
               }
-              className="custom-select"
-              custom
```
tyxchen, line 500: Missing these two attributes for custom styling.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-                  <li key={msg}>{ERRORS[msg]}</li>
-                ))}
-              </ul>
-            </Form.Control.Feedback>
-          </Col>
-        </Form.Group>
-      )}
-
-      <Form.Group as={Row} controlId="profileNewPassword">
+      {(props.user !== undefined && !auth.user?.is_editor) ||
```
tyxchen, line 246: Should use `auth.isEditor()`

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+        (props.user?.username === auth.user?.username && (
```
tyxchen, line 247: Sidenote, but we should probably remove this condition and force editors to enter their password even when changing/resetting other users' passwords.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-                    <FontAwesomeIcon icon={showPassword ? faEyeSlash : faEye} />
-                  </Button>
-                </InputGroup.Append>
-              </InputGroup>
-              <Form.Control
-                type="hidden"
-                isInvalid={
-                  state.errors.password && state.errors.password.length > 0
+        <Col sm={2}>
+          <Button
```
tyxchen, line 280: The button should be a secondary button, not a primary one.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-            </Col>
-          </Form.Row>
+                if (props.context.getValues().cur_password && !password) {
+                  return "USER.PASSWORD.NEW_REQUIRED";
+                }
+              },
+            })}
+          />
+        </Col>
+        <Col sm={1}>
```
tyxchen, line 314: I'm against putting the button in its own column; it makes for some weird alignment.

image

Would much rather the integrated-with-text-field look; it's pretty common by now, so people know what to expect.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+          <Button
```
tyxchen, line 315: The button should have clearly distinguishable on/off styles, and should be a secondary/tertiary button instead of a primary one.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-    if (
-      (Object.keys(state.changedUser) as Array<keyof ChangedUser>).every(
-        (k) =>
-          state.changedUser[k] === undefined ||
-          (k === "is_editor" && state.changedUser[k] === user?.is_editor)
-      )
-    ) {
-      return Promise.resolve();
+    if (props.onSubmit) {
+      await props.onSubmit(vals);
```
tyxchen, line 110: onSubmit has no way of knowing if form submission failed or not. This seems like an oversight.

src/admin/UserList.tsx
```diff
@@ -1,27 +1,77 @@
...
+        <Modal.Title>{props.user ? "Edit User" : "Create User"}</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <ProfileForm
+          context={context}
+          formId="editUserForm"
+          user={props.user}
+          onSubmit={async () => {
+            props.refreshTable();
+            props.setShow(false);
```
tyxchen, line 50: None of these should run if the submission was unsuccessful.

As well, I'd advise that the modal stay open even on a successful submit, so that we don't have to keep reopening it when creating multiple users.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-
-      {state.successMessage.length > 0 && (
-        <Alert
-          variant="success"
-          onClose={() => dispatch({ type: "set success message", message: "" })}
-          dismissible
-        >
-          {state.successMessage}
-        </Alert>
-      )}
```
tyxchen, line 372: These alerts should be kept in, so that the user knows if submission was successful or not.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
+            onClick={() => {
+              props.context.setValue("password", nanoid());
+              // set the password visible so you can see what you get
+              setShowPassword(true);
+            }}
+          >
+            Generate
+          </Button>
+        </Col>
+        <Col sm={7}>
```
tyxchen, line 291: Why is the generate password button separated into its own column?

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +23,315 @@ interface ChangedUser {
...
-        isChecking: false,
-        changedUser: { ...state.changedUser, ...action.data },
-        errors: { ...state.errors, ...action.errors },
-      };
-  }
-  return state;
+export const useProfileForm = (user?: User) => {
+  return useForm<ProfileFormVals>({
+    mode: "onBlur",
+    reValidateMode: "onBlur",
```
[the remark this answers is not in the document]
tyxchen: Alternatively, we keep validation onBlur for everything but username and password --- on these fields, we can write `onChange={() => triggerValidation(...)}`.

src/admin/UserList.tsx
```diff
@@ -1,27 +1,77 @@
...
 }
 
-type Action =
-  { type: 'set edit user', data: User | undefined } |
-  { type: 'show edit user', data: boolean } |
-  { type: 'show create user', data: boolean };
+const UserEditModal: React.FC<UserEditModalProps> = (
+  props: UserEditModalProps
+) => {
+  const context = useProfileForm(props.user);
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/e2888f5f4cae9180fa9b48a3dd2cf8ddf62e0d4c

src/admin/UserList.tsx
```diff
@@ -1,27 +1,77 @@
...
+          variant="secondary"
+          onClick={() => {
+            props.setShow(false);
+          }}
+        >
+          Close without saving
+        </Button>
+        <Button
+          type="submit"
+          disabled={formState.isSubmitting}
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/658917ce68dc06075c74919787a1e4d6412d1640

src/profile/ProfileForm.tsx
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/b58f6bd35e29d7b899b8e2b1260a8374156a313f

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/c57717f2d58a502ca8753aae95a11bdd6dbbe021

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/c57717f2d58a502ca8753aae95a11bdd6dbbe021

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/b0e8cb668c738162ad2cf8e6d917c120e042abc7

src/admin/UserList.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/22b3f1fd434e243f91d53ffd7c901162d0c89b3b

src/profile/ProfileForm.tsx
ktrieu: Probably going to just get rid of this check, it's confusing to have different behaviour for different users anyway.

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/53ece12a397300e594d32c8dc8ce2d7c9859a9ed

src/profile/ProfileForm.tsx
ktrieu: So we're asking editors to enter their own password next to the user's new password? That's going to be confusing unless we can communicate that somehow. Maybe change the label to "Your Password"?

src/profile/ProfileForm.tsx
ktrieu: My mastermind idea was that if an exception was thrown, it would catch and not call the onSubmit code, which would have worked if the call was actually inside the try block.

https://github.com/UWmathNEWS/slugline-web/pull/13/commits/0ae171358513d97053e8b78183be48cda056e5ed

src/admin/UserList.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/1fd4c2e2d1cc3cd580cf518d1928501bee99dc2d

To make that use case work, I also reset the form when a create completes. But it feels kind of... weird since it tells you that you created a user whose data is no longer there. Let me know what you think.

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/e85313c02adcbb5620a46d8d78d4a455ce6c3f19

src/profile/ProfileForm.tsx
tyxchen: Hmm, that's a good point. Or have something like GitHub where a modal pops up and you have to enter your own password when submitting?

src/admin/UserList.tsx
tyxchen: The ideal behaviour here would be:

- User fills in info, and submits
- Submission is successful
- Form clears, and an alert states that the previous submission was successful
- repeat

The key is that a notification is given on success. Seeing as that's been implemented, I think this is fine.

src/profile/ProfileForm.tsx
```diff
@@ -113,242 +130,270 @@ export const ProfileFormConsumer: React.FC<ProfileConsumerFormProps> = (
...
+              name="first_name"
+              placeholder="First Name"
+              ref={props.context.register({
+                maxLength: {
+                  value: 30,
+                  message: "USER.FIRST_NAME.TOO_LONG.30",
+                },
+                required: true,
+              })}
+            />
```
tyxchen, line 215: On further testing, I'm able to just tab over this field without it triggering and I'm not sure why. I think it has something to do with validation being onChange now (ugh). Here's my stab at a possible solution:

- Change validation back to being onBlur
- For only the username and password fields, force validation onChange

This way all fields are validated even if they're just tabbed over, and we still keep the live validation for the two fields that need it most.

ktrieu:
> On another note, I'm on the fence about placeholders --- should we bring them back, or is the form fine without them?

I don't really think we need them unless they're conveying some information the labels can't. So, in the name field, it's hard to tell what exactly is needed from just the name label (if you see an actual first and last name in the field, you should figure it our pretty fast). In that case, we need placeholders.

If its just a regular field, no, I don't think we need them, especially when they'll be exact duplicates of the labels, which looks kind of silly.

src/profile/ProfileForm.tsx
ktrieu: This might just fix a bug I've been struggling with too, so I'll keep you posted.

ktrieu: Well, it didn't fix my bug, but you can no longer tab past the first name field without being harassed about it.

https://github.com/UWmathNEWS/slugline-web/pull/13/commits/49fe676c2588c88a1f3379dbfa302c64c7ecc1a9

src/profile/ProfileForm.tsx
ktrieu: A modal inside another modal might be a bit much. I think a good way to do it would be to separate the new password and current password fields, and phrase the "current password" field more like "confirm your password".

And maybe some conditional rendering to only show it when you've entered a new password.

tyxchen: Yeah, I like that idea. Could look something like this:

image

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,387 @@ interface ChangedUser {
...
+                props.context.setValue("password", nanoid());
+                // set the password visible so you can see what you get
+                setShowPassword(true);
+              }}
+            >
+              Generate
+            </Button>
+          </Col>
+          <Col sm={8}>
+            <Field
```
tyxchen, line 347: Extremely minor, but the password field should be marked as required.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,387 @@ interface ChangedUser {
...
+                required: "USER.REQUIRED.first_name",
+              })}
+            />
+          </Col>
+          <Col sm={5}>
+            <Field
+              errors={props.context.errors}
+              type="text"
+              name="last_name"
+              placeholder="Last Name"
```
tyxchen, line 225: Minor: since this is optional, the placeholder should mention that.

src/profile/ProfileForm.tsx
ktrieu: Are we requiring password entry for changing any field? I assumed it was just for password changing.

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/c7182f99e122a2b497663f642b5ab8d7f63fa618

src/profile/ProfileForm.tsx
ktrieu: Just realized what you did with the nested row thing.
https://github.com/UWmathNEWS/slugline-web/pull/13/commits/f0dfc74467856918a8f3bbf4632d82efb7a04b51

src/profile/ProfileForm.tsx
tyxchen: Ack. Should've clarified --- it should only be required when creating a user.

ktrieu: Oh, that makes sense yeah.

ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/6ff2b84ef452f865cbd12226bf238a1ade257d49

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,385 @@ interface ChangedUser {
...
+            <Form.Row>
+              <Col sm="auto">
+                <Button
+                  className="w-100"
+                  variant="secondary"
+                  onClick={() => {
+                    props.context.setValue("password", nanoid());
+                    // set the password visible so you can see what you get
+                    setShowPassword(true);
+                  }}
```
tyxchen, line 319: When you generate a password, it doesn't trigger revalidation of the field; you have to change focus, then the onBlur validation kicks in: https://streamable.com/53ngus

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,385 @@ interface ChangedUser {
...
+  useEffect(() => {
+    register({
+      name: "is_editor",
+    });
+  }, [register]);
+
+  // we need a new password if we're creating a new user
+  const newPasswordRequired = props.user === undefined;
+  const passwordConfirmRequired =
+    newPasswordRequired || props.context.getValues().password;
```
tyxchen, line 100: This is getting convoluted, but: it would be ideal to not have to enter a password every time we create a new user, since that's normally done in bulk. Perhaps we can force a confirmation of password when creating a user only if we are making the user an editor (so `newPasswordRequired && props.context.getValues().is_editor`).

Similarly, it's probably best if we have to confirm password upon changing a user's role, to prevent power abuses (so `props.user && props.context.getValues().is_editor != props.user.is_editor`).

ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/aee2ca6a02de8bf4b25d42a4e79d85f08d671ea0
Are we enforcing these rules on the backend? If we're not, I'll add those in now.

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/07fb0662c68d0190ccfafe88acfde9fc6d2d29bf

src/profile/ProfileForm.tsx
tyxchen: > (so `props.user && props.context.getValues().is_editor != props.user.is_editor`)

The `props.user` check is still needed, since `false !== undefined` and so the password confirm dialog still shows when creating a bog-standard user.

tyxchen: > Are we enforcing these rules on the backend? If we're not, I'll add those in now.

That's something we should also do, yes.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,405 @@ interface ChangedUser {
...
-          }
-        >
-          {state.isLoading ? "Saving..." : "Save"}
-        </Button>
-      )}
-    </Form>
+            </Form.Group>
+          </>
+        )}
+        <NonFieldErrors errors={generalErrors || []} />
```
tyxchen, line 407: General errors should be moved to the top of the form.

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,405 @@ interface ChangedUser {
...
+          message: error,
+        })) || []
+      );
+      props.context.setError(
+        apiErrors.password?.map((error) => ({
+          name: "password",
+          type: "server_error",
+          message: error,
+        })) || []
+      );
```
tyxchen, line 163: Can we make this more general, seeing as server errors can be for any field (and can also be general, with the `detail` field only)? E.g.

```js
for (const [name, errArray] of apiErrors.entries()) {
  if (name === "detail") continue;
  props.context.setError(
    errArray.map((message) => ({
      name,
      type: "server_error",
      message
    })) || []
  );
}
```

src/profile/ProfileForm.tsx
```diff
@@ -37,585 +25,405 @@ interface ChangedUser {
...
-          state.changedUser,
-          user === auth.user
-        )
-    ).then(
-      () => {
-        dispatch({ type: "done loading success" });
-      },
-      (err: UserAPIError | string) => {
-        dispatch({ type: "done loading error", errors: err });
+      setGeneralErrors(undefined);
```
tyxchen, line 144: We should also reset showPassword to be false here, so users don't get a nasty surprise.

src/admin/UserList.tsx
```diff
@@ -49,186 +41,159 @@ const columns: Column<User>[] = [
...
         </Button>
-        <Button variant="outline-danger" onClick={() => {
-          if (tableBagRef.current) {
-            tableBagRef.current.executeAction("Delete");
-          }
-        }}>Delete user</Button>
         <Button
           type="submit"
-          disabled={isLoading || editUserErrors}
-          className="ml-auto"
```
tyxchen, line 179: Submit button should keep this class, so that the "Cancel" and "Save" actions are physically (and socially) distanced from each other.

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/b614bfaf1fda88049bccda144a4f1230dcd1743e

src/admin/UserList.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/37baf6c8a17f29cdfce3cf4aed9a1aa1a50270c3

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/24108681801fdd0e22ee5fd2198b3ffaac4bca46

src/profile/ProfileForm.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/13/commits/0a4e2cec367de918e069b348696f67eeacd0c269

ktrieu:
Alright I think that's it. You have full right to merge this if I'm not awake.

PR #16, opened 2020-05-10
Add useful test

vincentmacri:
With this useful test, we can know immediately if your changes broke everything.


src/App.test.tsx
```diff
@@ -4,6 +4,5 @@ import App from "./App";
 
 test("renders learn react link", () => {
```
ktrieu, line 5: Should change the default test's name, too.

PR #17, opened 2020-05-10
Font fixes

### LINKED ISSUE #14: Missing fonts
Currently the project relies on a font file that we can't redistribute.

Instead, we should distribute some open font and have a config option to change the font that is used in production.

This will allow easy testing while still being able to use whatever font we want in production.

vincentmacri:
Fixes #14


tyxchen:
I believe the long term plan is to use Typekit for font delivery. I haven't thought of exactly how we would do that, but to my knowledge, it's not possible to use `@font-face` with Typekit owing to how they generate and serve their fonts.

By the way, how would the license for the font config files work, as well as the variable files? I'm thinking we might have to license them permissively, because we'll most definitely have to edit them on production, and there's a chance those edits may include secrets (see: Typekit) that we cannot release publicly.

tyxchen:
I should also note that in mockups, I've used a combination of Merriweather/Overpass, though I believe the closest combo to our brand is Spectral/Overpass.

vincentmacri:
> I should also note that in mockups, I've used a combination of Merriweather/Overpass, though I believe the closest combo to our brand is Spectral/Overpass.

I've changed the fonts to this.

src/_slugline_vars.scss
```diff
@@ -2,9 +2,7 @@
 
 $primary: #f2c0d9;
 
-// Use our custom fonts
-$body-ff: "Joanna Nova", serif;
-$headings-ff: "DIN Next", sans-serif;
+@import "slugline_fonts";
```
ktrieu, line 5: This import (and any SCSS that generates a CSS rule), needs to be moved to slugline.scss, or any component that imports this variable file will create a copy of the `font-face` selectors.

slugline.scss is imported only once, so any top-level rules should go there.

src/slugline.scss
```diff
@@ -2,6 +2,17 @@
 
 @import "~bootstrap/scss/bootstrap.scss";
 
+@font-face {
+  font-family: "Body Font";
+  src: url("https://fonts.googleapis.com/css2?family=Merriweather&display=swap");
+}
+
+@font-face {
+  font-family: "Headings Font";
```
ktrieu, line 11: This doesn't match the `font-family` in the variables.

src/slugline.scss
```diff
@@ -1,26 +1,27 @@
...
 
-@import "slugline_vars";
-
-@import "~bootstrap/scss/bootstrap.scss";
 
 // Convenient class for mathnews logo
 .mathnews-logo::before {
   content: "math";
-  font-family: Joanna Nova;
+  font-family: Body Font;
```
tyxchen, line 19: This rule (and the one below) should be changed to use variables.

src/_slugline_vars.scss
```diff
@@ -3,8 +3,8 @@
 $primary: #f2c0d9;
 
 // Use our custom fonts
-$body-ff: "Joanna Nova", serif;
-$headings-ff: "DIN Next", sans-serif;
+$body-ff: "Body Font", serif;
+$headings-ff: "Serif Font", sans-serif;
```
tyxchen, line 7: Should be `"Headings Font"`, otherwise font-face won't work

src/slugline.scss
```diff
@@ -1,26 +1,28 @@
...
+  font-family: "Body Font";
+  src: url("https://fonts.googleapis.com/css2?family=Merriweather&display=swap");
 }
 
 @font-face {
-  font-family: "Joanna Nova";
-  src: url("./fonts/joanna_nova.woff") format("woff");
+  font-family: "Headings Font";
+  src: url('https://fonts.googleapis.com/css2?family=Overpass&display=swap');
 }
```
tyxchen, line 13: I realize this is confusing, but these font face rules should really be put in a separate file for configuration. This second file would then be imported only once, in `slugline.scss`, while the variables defining the font names stay in `_slugline_vars` and are thus visible to the entire project.

src/_slugline_fonts.scss
```diff
@@ -0,0 +1,9 @@
+@font-face {
+  font-family: "Body Font";
+  src: url("https://fonts.googleapis.com/css2?family=Merriweather&display=swap");
+}
+
+@font-face {
+  font-family: "Headings Font";
+  src: url('https://fonts.googleapis.com/css2?family=Overpass&display=swap');
```
tyxchen, line 8: nit: replace single quotes with double quotes

vincentmacri: Missed that one, thanks.

PR #20, opened 2020-05-11
Introduce tests (part 1)

tyxchen:
This PR introduces tests for auth/PrivateRoute and api/*, and partial coverage for App. It also introduces fixes for bugs uncovered in those components while testing.


vincentmacri:
I would recommend merging in each test as you finish it, because getting them into the repo faster means we can spend less time manually testing other changes.

tyxchen:
> I would recommend merging in each test as you finish it, because getting them into the repo faster means we can spend less time manually testing other changes.

I'm planning to merge in a few tests at a time. Gonna finish this one up tonight.

src/auth/PrivateRoute.tsx
```diff
@@ -95,12 +97,14 @@ const PrivateRouteWrapper: React.FC<{
...
+        return <Error404 />;
       }
     } else {
       return <>
-        {props.fallback ?? <h1>Loading...</h1>}
-        <div key="children" className="d-none">{props.children}</div>
+        {props.fallback ?? <Spinner animation="border" role="status">
+          <span className="sr-only">Loading...</span>
+        </Spinner>}
+        <div key="children" className="d-none" aria-hidden="true">{props.children}</div>
```
ktrieu, line 107: This is maybe an admission that I didn't notice this when we initially reviewed, but why are we hiding the children here instead of just not rendering them?

src/index.tsx
```diff
@@ -4,9 +4,22 @@ import "./index.css";
...
 
 // If you want your app to work offline and load faster, you can change
 // unregister() to register() below. Note this comes with some pitfalls.
 // Learn more about service workers: https://bit.ly/CRA-PWA
 serviceWorker.unregister();
+
+render(App);
+
+// @ts-ignore
+if (module.hot) {
```
ktrieu, line 19: What's this section for? Guessing its trying to fix a hot-reload issue?

Whatever the reason, there should be a comment here explaining it, since its ignoring compiler errors and changing a pretty common React pattern.

src/auth/__tests__/PrivateRoute.unit.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+            success: false,
+            error: {
+              detail: ["Error thrown in testing"]
+            }
+          }
+        }
+      });
+    });
+
+    expect(getByText("Error thrown in testing")).toBeInTheDocument();
```
ktrieu, line 201: I recognize we're doing this already, but do we want to be directly outputting auth errors into the document? We should run them through the errors map first.

src/setupTests.ts
```diff
@@ -3,3 +3,25 @@
 // expect(element).toHaveTextContent(/react/i)
```
ktrieu, line 3: Can probably get rid of this comment now.

src/auth/PrivateRoute.tsx
```diff
@@ -95,12 +97,14 @@ const PrivateRouteWrapper: React.FC<{
...
-        return <Redirect to="/login/" />;
+        return <Error404 />;
       }
     } else {
       return <>
-        {props.fallback ?? <h1>Loading...</h1>}
-        <div key="children" className="d-none">{props.children}</div>
+        {props.fallback ?? <Spinner animation="border" role="status">
+          <span className="sr-only">Loading...</span>
+        </Spinner>}
```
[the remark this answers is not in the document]
tyxchen: I've been thinking the same thing, but was going to leave it as being outside of the scope of this PR.

src/auth/PrivateRoute.tsx
tyxchen: Performance --- we can do a bit of mounting/rendering/API calls while the spinner is visible.

src/index.tsx
tyxchen: `hot` isn't defined on `module` types, so Typescript complains.

src/auth/__tests__/PrivateRoute.unit.test.tsx
tyxchen: I believe they're already being run through the errors map. This is just testing that the specific error message returned is being displayed.

src/setupTests.ts
tyxchen: cd2e1af

src/index.tsx
ktrieu: Some Googling indicates we can install `@types/webpack-env` for that.  Dunno if you want to bother.

src/auth/__tests__/PrivateRoute.unit.test.tsx
ktrieu: Unless I'm missing something, we're just outputting them directly:

```
return <>
  {state.errors.map(err => <Alert key={err} variant="danger">{err}</Alert>)}
</>;
```

tyxchen: Oh, I see what you mean. Yeah, that's a good call; I'll also test to make sure errors are being passed through the map on all of the tests.

tyxchen: 52141354

src/index.tsx
tyxchen: For a single section, I don't think it's worth installing a package for. I've added a comment explaining the situation instead: fcaa7cf

PR #24, opened 2020-05-14
Run npm ci instead of npm install for ci

vincentmacri:
This PR implements this: https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -36,7 +36,7 @@ jobs:
 
     # Runs a single command using the runners shell
     - name: Install dependencies
-      run: npm install
+      run: npm ci
 
     - name: Run tests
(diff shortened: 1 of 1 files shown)


vincentmacri:
Based on the test that just ran on this PR, this PR seems to cut CI time by roughly half.

tyxchen:
This is interesting!

Something I've noticed (and tried to fix yesterday, to no avail) is that the package cache doesn't appear to be working --- `npm install` is still installing a large number of packages, when theoretically with a cache we should see something like "added 0 packages". This persists with `npm ci`, which (1) tells me that there's definitely something up with the cache, and (2) makes me wonder how much faster this would be if the cache actually was working.

PR #25, opened 2020-05-14
Fix issues in RichTable

tyxchen:
This fixes issues with pagination in RichTable, and introduces some enhancements and abstractions to make it easier to extend and use.

~~TODO: add tests~~ 2020-05-18: Will make tests part of a later PR, this should be merged in sooner rather than later

---

Blocked by: none
Blocks: #33


src/shared/components/RichTable.tsx
```diff
@@ -593,6 +627,73 @@ export const RichTable = <D extends object = {}>(config: RichTableProps<D>) => {
...
+              onKeyDown={({ key, currentTarget }: React.KeyboardEvent<HTMLInputElement>) => {
+                if (key === "Enter") {
+                  const requestedPage = Math.max(1, Math.min(parseInt(currentTarget.value), numPages)) || page;
+                  setPage(requestedPage);
+                  setDisplayPageSelector(false);
+                } else if (key === "Escape") {
+                  setDisplayPageSelector(false);
+                }
+              }}
+              onBlur={() => { setDisplayPageSelector(false) }}
```
ktrieu, line 657: This might be just me, but I expected blurring the selector to also set the page.

src/shared/components/RichTable.scss
```diff
@@ -32,6 +39,10 @@
   &_table {
     table-layout: fixed;
     margin-bottom: 0;
+
+    @include media-breakpoint-down(sm) {
+      width: auto;
```
ktrieu, line 44: This makes tables with only a few columns not take up all the space on small screen sizes:
screenshot

Tried setting it to 100%, but then tables with lots of columns get crushed and all the text has to break. Think in the future, we might have to specify "important columns" and then drop the rest on small screen sizes.

ktrieu:
Found a bug where if you disable the network and keep paging in the error state, the table is permanently broken, even when the network is re-enabled.

src/shared/components/RichTable.tsx
tyxchen: Took longer than I expected, but I figured out how to have an ideal submit-on-blur: 30eb79f

src/shared/components/RichTable.scss
tyxchen: Fixed in 27ad7b3

tyxchen:
> Found a bug where if you disable the network and keep paging in the error state, the table is permanently broken, even when the network is re-enabled.

So it turns out that we never reset error state in useApiGet, even after a successful response. I've since fixed this in fde2b32.

ktrieu:
> > Found a bug where if you disable the network and keep paging in the error state, the table is permanently broken, even when the network is re-enabled.
> 
> So it turns out that we never reset error state in useApiGet, even after a successful response. I've since fixed this in [fde2b32](https://github.com/UWmathNEWS/slugline-web/commit/fde2b32a9aa712b1c3b38d39ddc1c12ed129f39d).

Oops, my fault then. I'll add tests for this case in the new API branch as well.

src/shared/components/RichTable.tsx
ktrieu: Hate to nitpick some more on this, but the behaviour of the input box replacing your input with valid input as you type is kinda weird, IMO. It should only do that on submit.

tyxchen: 7e8a601

PR #26, opened 2020-05-14
Optimize CI

vincentmacri:
This PR fixes CI caching once and for all.

---

Blocked by: ~~#29~~
Blocks: none

PR #29, opened 2020-05-18
Introduce tests (part 2)

tyxchen:
Introduces tests for `shared/errors`, `shared/hooks/*`, and `shared/form/*`.

---

Blocked by: none
Blocks: #26

### DIFF
--- a/src/shared/__tests__/errors.unit.test.ts
+++ b/src/shared/__tests__/errors.unit.test.ts
@@ -83,10 +83,10 @@ describe("ERRORS", () => {
     }
   });
 
-  describe("USER.REQUIRED", () => {
+  describe("REQUIRED", () => {
     it("uses correct a/an form", () => {
-      expect(ERRORS.USER.REQUIRED.email).toMatch(/\ban\b/);
-      expect(ERRORS.USER.REQUIRED.password).toMatch(/\ba\b/);
+      expect(ERRORS.REQUIRED.email).toMatch(/\ban\b/);
+      expect(ERRORS.REQUIRED.password).toMatch(/\ba\b/);
     })
   });
 });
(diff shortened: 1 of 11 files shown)


ktrieu:
Looks pretty much fine, though I'm unsure about throwing an error on an invalid error map. What I've seen with `react-18n` is it displays the key as a fallback instead of throwing an exception and crashing the app.

tyxchen:
I figure that once we get all the testing done, such crashes will be caught in integration testing. So I'd rather throw the error than silently fail.

ktrieu:
Alright, sounds fair to me.

src/shared/__tests__/errors.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  it("rejects adding additional errors to nested levels", () => {
+    try {
+      ERRORS.__TEST.__ASDFUHFAI = "?";
+      expect("Error was not thrown by ErrorsProxy").toBe(false);
+    } catch (e) {
+      expect(e.message.includes("Error was not thrown by ErrorsProxy")).toBe(false);
+    }
+  });
+
+  describe("USER.REQUIRED", () => {
```
ktrieu, line 86: We'll probably end up using the required form in more places than the user form, so we should move it to the top-level.

tyxchen: The way I structured this test suite (and all my test suites) is that top-level test groups all correspond to a module export. We then test functionality related to that export within the test group. If a method or part of an export requires more granular testing (e.g. here with `USER.REQUIRED`), we declare a nested test group. This way, the organization of test suites follow the organization of their corresponding modules.

ktrieu: I probably worded that weird. What I meant is that in the error map, we should move `USER.REQUIRED` into just `REQUIRED`, seeing as the text is pretty generic and doesn't reference the user form at all.

tyxchen: Ahh, I see what you mean. Moved and deprecated in 02c0bea.

PR #30, opened 2020-05-18
Update bootstrap

tyxchen:
This PR updates the version of react-bootstrap used from 1.0.0-beta.17 to 1.0.1 (latest).


src/profile/ProfileForm.tsx
```diff
@@ -300,7 +300,6 @@ export const ProfileFormConsumer: React.FC<ProfileConsumerFormProps> = (
                 value={
                   props.context.getValues().is_editor ? "editor" : "contributor"
                 }
-                className="custom-select"
```
ktrieu, line 303: I'm still getting an error in the console: `Warning: Received 'true' for a non-boolean attribute 'custom'.`, which I thought this would fix.
Are you getting the same thing?

tyxchen: Does `npm list react-bootstrap` show it at 1.0.1? If not, you'll have to run `npm install` --- though that'll break your other branches.

ktrieu: Yeah, it says 1.0.1. That's weird if you're not getting it.

tyxchen: Hmm, try as I might, I don't get that issue. Try a clean rebuild?

ktrieu: Ok, now it's gone. Guess something didn't refresh properly when I reinstalled the deps. False alarm.

PR #31, opened 2020-05-20
Badges

vincentmacri:
This PR fixes the badge links and also runs the security checks action on master daily.


tyxchen:
Nice!

Curious, but can we do something similar to npm audit for the backend too?

vincentmacri:
> Nice!
> 
> Curious, but can we do something similar to npm audit for the backend too?

I'll look into that next.

PR #34, opened 2020-05-22
Modernize AuthProvider

### LINKED ISSUE #7: AuthProvider changes whenever auth.check is called
See https://github.com/UWmathNEWS/slugline-web/pull/6#issuecomment-620988218 for context.

Possible fixes would be to move `readyPromise` into a ref instead of state, or as a module-level variable. Both could have unintended consequences, however.

tyxchen:
This PR adds tests for AuthProvider and fixes #7.

---

Blocked by: none
Blocks: #33


src/auth/__tests__/AuthProvider.unit.test.tsx
```diff
@@ -0,0 +1,428 @@
...
+      expect(auth).not.toBe(result.current);
+
+      auth = result.current;
+      expect(auth.user).not.toEqual(testUser);
+    });
+
+    it("handles unsuccessful requests by throwing an error", async () => {
+      let checkPromise: Promise<any>;
+
+      checkPromise = auth.check(true)!; // if the force flag is true then check must return a promise
```
ktrieu, line 232: Not quite sure what's happening here: according to the types, `check` always returns a `Promise<void>`.

tyxchen: That was before I changed the types so that that would be true. I've since fixed the test.

PR #36, opened 2020-05-26
Size limit test

vincentmacri:
This PR adds a size limit check to the CI.

---

Blocked by: #37 
Blocks: none

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -10,8 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CI_JOB_NUMBER: 1
-      steps:
-        - uses: actions/checkout@v1
-        - uses: andresz1/size-limit-action@v1.3.0
-          with:
-            github_token: ${{ secrets.GITHUB_TOKEN }}
+    step:
+      - uses: actions/checkout@v1
+      - uses: andresz1/size-limit-action@v1.3.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -20,3 +20,4 @@ jobs:
     - uses: andresz1/size-limit-action@v1.3.0
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        skip_step: install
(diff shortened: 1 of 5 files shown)


vincentmacri:
@tyxchen it's broke, ideas?

tyxchen:
@vincentmacri So it looks like the workflow is actually mostly working as intended. The only reason it's failing is because it's testing the size of master, but master doesn't have size-limit installed, so the scripts fail. In the meantime, here's my suggestions to get it ready for when we do that:

- Reset to 8ae2509, which is the last commit that provides a working workflow.
- Add size-limit and @size-limit/preset-app under devDependencies in package.json. This way, when we merge this PR into master, future PRs will run this check correctly.
- Remove the step to install size-limit in the workflow. The previous step covers this.

package.json
```diff
@@ -56,6 +62,7 @@
   },
   "proxy": "http://localhost:8000",
   "devDependencies": {
+    "@size-limit/preset-app": "^4.5.0",
```
tyxchen, line 65: You still need to specify size-limit as a dependency here.

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches:
+      - master
+
+jobs:
+  size:
+    runs-on: ubuntu-16.04
+    env:
+      CI_JOB_NUMBER: 1
+
+    steps:
```
tyxchen, line 14: Is it possible to add a restore-from-cache step?

vincentmacri: Probably, but given the pains I've had getting size-limit to work, I want to add the restore from cache after I get size-limit working. Then I only have to debug one thing at a time.

package.json
vincentmacri: Ah, so that's why it wasn't working. I can't read.

vincentmacri:
I have no idea why it's giving the error my message. My only assumption is because the test hasn't run on master yet.

vincentmacri:
As for caching, I will do that after this is merged because I don't want to be debugging multiple things at a time. Once this is working I'll add caching.

tyxchen:
> I have no idea why it's giving the error message. My only assumption is because the test hasn't run on master yet.

Your assumption is correct, size-limit is working great on this branch. We should be able to test this on #38.

> As for caching, I will do that after this is merged because I don't want to be debugging multiple things at a time. Once this is working I'll add caching.

Sounds good to me.

PR #43, opened 2020-06-14
Create LICENSE and other nice things

### LINKED ISSUE #12: No LICENSE file
This project is missing a LICENSE file.

I would recommend the Affero General Public License (AGPL). The AGPL is essentially the same as the GPL with some additional clauses to enforce the spirit of free software when the software is run on a server.

vincentmacri:
Fixes #12.


vincentmacri:
@tyxchen I including wording in the README that some files may be under a different license. This is to deal with the issue of configuration files and such.

CONTRIBUTING.md
```diff
@@ -0,0 +1,2 @@
+# Contributing
+Submit a pull request or something.
```
vincentmacri, line 2: Should we have a contributor agreement?

README.md
```diff
@@ -9,5 +9,23 @@ The front-end component of `slugline`, math**NEWS**'s best and only React/Django
 - Grab a copy of the server [here](https://github.com/UWmathNEWS/slugline-api). 
 - By default, the server runs at port 8000, and the `proxy` setting in `package.json` is designed to redirect requests from the dev server to there. If you change the server port, change the port in `package.json` as well.
 
-## Contributing
-Submit a pull request or something.
+## License
+Except for files which specify otherwise, this application is licensed under the [GNU Affero General Public License](https://www.gnu.org/licenses/agpl-3.0.en.html).
+
+```
+<one line to give the program's name and a brief idea of what it does.>
```
vincentmacri, line 16: Not sure if we've settled on the name yet.

README.md
```diff
@@ -9,5 +9,23 @@ The front-end component of `slugline`, math**NEWS**'s best and only React/Django
...
+Copyright (C) 2020  <name of author>
```
vincentmacri, line 17: Who are the copyright holders?

README.md
```diff
@@ -9,5 +9,23 @@ The front-end component of `slugline`, math**NEWS**'s best and only React/Django
...
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
```
vincentmacri, line 22: Thoughts on any later version? I'm somewhat uncomfortable with releasing under a license I haven't seen, but also it should be easy to use this under AGPL v4 if it's every released, provided that v4 isn't bad.

CONTRIBUTING.md
tyxchen: IMO not necessary, the GPL is implicit agreement for all the scenarios I can think of (which aren't many).

@ktrieu thoughts?

README.md
tyxchen: I don't think we need this section in our README, considering we may have to go for per-file headers on this one.

README.md
tyxchen: Copyright is held by individual contributors for only the files they contributed to.

CONTRIBUTING.md
ktrieu: Yeah, I think we'll be OK without one for now.

ktrieu: Yeah, I think we'll be OK without one for now.

vincentmacri:
@tyxchen Thoughts on this: https://www.gnu.org/licenses/gpl-faq.html#WMS

tyxchen:
> @tyxchen Thoughts on this: https://www.gnu.org/licenses/gpl-faq.html#WMS

Yeah, this sounds like a good idea.

I think we should look into attaching the license paraphrase as a per-file header. This provides the following benefits:

1. Copyright is held by contributors/authors of individual files. This removes the need for a CLA.
2. We can enforce including mathNEWS as an author in our contributing guidelines. Here's an example of how we might implement that: https://github.com/appleseedhq/appleseed/wiki/Copyright-Notices
3. It makes it clear which files are licensed under AGPL, and which files are not.

The downsides are that it makes name changes and license upgrades (if we decide to go that route) a pain.

vincentmacri:
> > @tyxchen Thoughts on this: https://www.gnu.org/licenses/gpl-faq.html#WMS
> 
> Yeah, this sounds like a good idea.
> 
> I think we should look into attaching the license paraphrase as a per-file header. This provides the following benefits:
> 
>     1. Copyright is held by contributors/authors of individual files. This removes the need for a CLA.
> 
>     2. We can enforce including mathNEWS as an author in our contributing guidelines. Here's an example of how we might implement that: https://github.com/appleseedhq/appleseed/wiki/Copyright-Notices
> 
>     3. It makes it clear which files are licensed under AGPL, and which files are not.
> 
> 
> The downsides are that it makes name changes and license upgrades (if we decide to go that route) a pain.

For 1, what if there end up being a lot of contributors to one file?

vincentmacri:
@tyxchen because we don't live in a magical land where there are never any merge conflicts, I'd recommend that we don't add in the license headers with this commit, and instead add them as we change our files. Maybe in a week or two after merging this we can change any stragglers.

tyxchen:
At some point I can write a script to troll through all licensable files, extract the authors, and automatically add a license header.

vincentmacri:
> At some point I can write a script to troll through all licensable files, extract the authors, and automatically add a license header.

Allegedly GitHub actions for this exist, but at a certain point I don't really want to use third party things for everything.

vincentmacri:
@tyxchen we discussed the FSF's response offline, so I think this is ready.

To summarise for future people: the webpages are the output of the program and thus the AGPL doesn't apply to our website content. I feel the need to say this is not legal advice. This is not legal advice.

.github/PULL_REQUEST_TEMPLATE/devops.md
```diff
@@ -0,0 +1,10 @@
+---
+name: DevOps
+about: Propse changes to the DevOps pipeline
```
tyxchen, line 3: Propse --> propose

CONTRIBUTING.md
```diff
@@ -0,0 +1,30 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>.
+```
+We also include the following [exception](https://www.gnu.org/licenses/gpl-faq.en.html#WMS) to the AGPL for this software.
+```
+As a special exception to the AGPL, any HTML file which merely makes function calls to this code, and for that purpose includes it by reference shall be deemed a separate work for copyright law purposes. In addition, the copyright holders of this code give you permission to combine this code with free software libraries that are released under the GNU LGPL. You may copy and distribute such a system following the terms of the GNU AGPL for this code and the LGPL for the libraries. If you modify this code, you may extend this exception to your version of the code, but you are not obligated to do so. If you do not wish to do so, delete this exception statement from your version.
+```
```
tyxchen, line 30: Should the exception be placed in LICENSE?

vincentmacri: I thought we determined that we didn't need the exception actually.

vincentmacri:
@ktrieu I would like your review since this changes our license.

PR #48, opened 2020-06-15
Update size-limit-action

tyxchen:
This also undoes #47.

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -24,13 +24,13 @@ jobs:
         key:  ${{ runner.os }}-npm-cache-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
         restore-keys: |
           ${{ runner.os }}-npm-cache-${{ env.cache-name }}-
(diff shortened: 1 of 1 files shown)


.github/workflows/size-limit.yml
```diff
@@ -24,13 +24,13 @@ jobs:
...
           ${{ runner.os }}-npm-cache-${{ env.cache-name }}-
-          #${{ runner.os }}-npm-cache-
+          ${{ runner.os }}-npm-cache-
 
     - name: Install dependencies
       if: steps.npm-cache.outputs.cache-hit != 'true'
       run: npm install --no-audit --no-fund
 
-    - uses: andresz1/size-limit-action@v1.3.0
+    - uses: andresz1/size-limit-action@v1.3.2
```
vincentmacri, line 33: Isn't there @latest or something?

tyxchen: I'm not sure we want to go with @latest --- what if there are breaking changes? So I'd rather upgrade manually.

PR #49, opened 2020-06-15
Remove .vscode directory

ktrieu:
I swear I .gitignored this.


vincentmacri:
We should still document our coding style somewhere.

PR #52, opened 2020-06-16
Only track file sizes with size-limit

### LINKED ISSUE #51: size-limit timing is inconsistent
**Describe the bug**
The CI size-limit tool is inconsistent at measuring execution time.

**Expected behavior**
Execution time should not vary much for a tool intended to measure execution time.

**Fix**
Just don't bother tracking execution time. Instead just track loading time, which is based off file size only. This has the benefit of removing Chromium as a dependency.

vincentmacri:
This fixes #51.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -1,6 +1,6 @@
 [
   {
     "path": ["build/static/js/*.js", "build/static/css/*.css"],
-    "limit": "6 s"
+    "limit": "512 KB"
   }
 ]
--- a/package-lock.json
+++ b/package-lock.json
@@ -1423,39 +1423,6 @@
         }
       }
     },
-    "@size-limit/preset-app": {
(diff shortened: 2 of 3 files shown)


vincentmacri:
This fixed #51.

.size-limit.json
```diff
@@ -1,6 +1,6 @@
 [
   {
     "path": ["build/static/js/*.js", "build/static/css/*.css"],
-    "limit": "6 s"
+    "limit": "512 KB"
```
tyxchen, line 4: In the future, I'd honestly like to see this go lower --- I think we can live with a ~100 KB limit for our public-facing code and 300-400 KB for all files (including vendor packages).

vincentmacri: For sure, I just wanted a conservative estimate initially.

PR #53, opened 2020-06-17
Add code coverage reporting and improve CI

### LINKED ISSUE #50: Add code coverage tracking through CI
**Is your feature request related to a problem? Please describe.**
We don't currently have a way of tracking code coverage.

**Describe the solution you'd like**
I'd like CI to tell us the code coverage of the main branch and of PRs.

vincentmacri:
This fixes #50.

It also cleans up the naming of the different CI workflows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
       - master
 
 jobs:
-  size:
+  build:
     runs-on: ubuntu-16.04
     env:
       CI_JOB_NUMBER: 1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: "Size limit"
+name: "Build"
 
 on:
   pull_request:
@@ -18,7 +18,7 @@ jobs:
       id: npm-cache
       uses: actions/cache@v1
       env:
-        cache-name: size-limit
+        cache-name: build
       with:
         path: node_modules
         key:  ${{ runner.os }}-npm-cache-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -1,19 +1,19 @@
...
   build:
-    runs-on: ubuntu-16.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, ubuntu-latest, windows-latest, macos-latest]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v1
```
tyxchen, line 16: Curious why the downgrade here?

vincentmacri: Don't know, that was an accident.

.github/workflows/test.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Install dependencies
+      if: steps.npm-cache.outputs.cache-hit != 'true'
+      run: npm install --no-audit --no-fund
+    
+    - name: Check type errors
+      run: npx tsc --noEmit
+
+    - name: Run tests
+      if: matrix.os != 'ubuntu-16.04'
+      run: npm test -- --coverage
```
tyxchen, line 42: Do we need to generate a coverage report for non-16.04 pipelines?

vincentmacri: No, it seems to just upload the same thing.

I think I could easily get it to do code coverage for the other operating systems, but doing a coverage test is slower, and I only care about non-16.04 for making sure devs don't have trouble.

In theory there could be something that only happens in certain operating systems. I could try doing coverage for all if you want.

vincentmacri: Oh I see what you meant now. That's a typo. Will fix.

PR #59, opened 2020-06-19
Add tests for RichTable

tyxchen:
This should hopefully resolve some of our coverage issues.

Also improves coverage for Visor, Loader, and shared/hooks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,4 +48,4 @@ jobs:
     - uses: codecov/codecov-action@v1
       if: matrix.os == 'ubuntu-16.04'
       with:
-        file: coverage/coverage-final.json
+        file: coverage/lcov.info
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "jest": {
-    "coverageReporters": ["json", "text"]
+    "coverageReporters": ["json", "text", "lcov"]
   },
   "dependencies": {
     "@babel/cli": "^7.10.1",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "private": true,
   "jest": {
     "coverageReporters": [
-      "json",
       "text",
       "lcov"
     ]
--- a/package.json
+++ b/package.json
@@ -3,7 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "jest": {
-    "coverageReporters": ["json", "text", "lcov"]
+    "coverageReporters": [
+      "json",
+      "text",
+      "lcov"
+    ]
   },
   "dependencies": {
     "@babel/cli": "^7.10.1",
@@ -46,7 +50,7 @@
     "react-helmet": "^6.0.0",
     "react-hook-form": "^5.6.1",
     "react-router-dom": "^5.1.2",
-    "react-scripts": "^3.4.0",
+    "react-scripts": "^3.4.1",
     "serialize-javascript": "^3.1.0",
     "slate": "^0.57.1",
     "slate-react": "^0.57.1",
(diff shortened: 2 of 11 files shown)


tyxchen:
@vincentmacri So it looks like the negative changes in coverage might be something to live with as we slowly add more and more tests and clean up/fix code in doing so, since many of the lines that changed to being marked as untested were never properly tested to begin with.

tyxchen:
@vincentmacri Also, nice job on the size-limit action PR --- not being spammed with comments with every push is a huge improvement :smiley:

.github/workflows/test.yml
```diff
@@ -48,4 +48,4 @@ jobs:
     - uses: codecov/codecov-action@v1
       if: matrix.os == 'ubuntu-16.04'
       with:
-        file: coverage/coverage-final.json
+        file: coverage/lcov.info
```
vincentmacri, line 51: Why this change?

package.json
```diff
@@ -5,7 +5,8 @@
   "jest": {
     "coverageReporters": [
       "json",
```
vincentmacri, line 7: If you're using lcov do you also need json?

.github/workflows/test.yml
vincentmacri: Nevermind, saw your commit message explaining.

src/shared/components/__tests__/RichTable.unit.test.tsx
```diff
@@ -0,0 +1,2092 @@
...
+      const { result, waitForNextUpdate } = renderHook(() =>
+        useRichTable<TestData>({
+          columns: [{ key: "id", header: "ID" }],
+          list: apiWithoutPagination,
+          paginated: false,
+        })
+      );
+      await waitForNextUpdate();
+
+      render(<div {...result.current.header.useRowProps()} />);
```
ktrieu, line 920: This test should have an `expect` in it for clarity.

src/shared/hooks.ts
```diff
@@ -80,16 +81,16 @@ export const useSSRData: {
...
-    transformer: (resp: TResp) => TData,
+    transformer: (resp: TResp) => TData
   ): UseSSRDataHook<TData>;
   <TData, TError extends APIError = APIError>(
     dataMethod: () => Promise<APIResponse<TData, TError>>,
     initialData: TData | undefined
   ): UseSSRDataHook<TData>;
-} = <TData, TResp, TArgs, TError>(
-  dataMethod: (args?: any) => any,
+} = <TData, TResp, TError>(
```
ktrieu, line 90: VSCode informs me that `TResp` and `TError` aren't being used either, so we can get rid of those too.

src/shared/components/RichTable.tsx
```diff
@@ -616,12 +646,14 @@ const useRichTable = <D extends object = {}>({
 const RichTablePagination = ({ bag }: { bag: RichTableBag<any> }) => {
   const { page, numPages, setPage, reqInfo, executeAction } = bag;
   const [newPage, setNewPage] = useState(page.toString());
-  // we need a ref to store the new page to get around the fake blur listener capturing outdated values of newPage
-  const newPageRef = useRef(page.toString());
-  const hasBlurListener = useRef(false);
+  // we need a ref to store the page to get around the fake blur listener capturing outdated values of page
+  const pageRef = useRef(page);
+  const fakeBlurListener = useRef<((e: MouseEvent) => void) | null>(null);
```
ktrieu, line 651: Why are we using a fake blur listener here? Whatever the reason, it should probably be documented.

tyxchen: Documentation added: 933e930

src/shared/hooks.ts
tyxchen: 2009067

src/shared/components/__tests__/RichTable.unit.test.tsx
tyxchen: 4b7a148

src/shared/components/RichTable.tsx
```diff
@@ -680,51 +683,56 @@ const RichTablePagination = ({ bag }: { bag: RichTableBag<any> }) => {
...
-              e.stopPropagation();
-            }}
-            onFocus={({
-              currentTarget,
-            }: React.FocusEvent<HTMLInputElement>) => {
-              currentTarget.select();
+            onFocus={() => {
+              inputRef.current?.select();
 
               // only submit if blur was the result of a click outside the control
```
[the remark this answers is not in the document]
tyxchen: 933e930

package.json
tyxchen: Good catch, I guess not. 9b6063c

PR #65, opened 2020-06-24
Fix race condition in PrivateRoute

### LINKED ISSUE #64: Certain private routes don't load properly in dev server
**Describe the bug**
When trying to access the profile or admin routes from either a direct link or via an HMR update with the default webpack dev server setup (NOT our custom SSR server), the routes get stuck in the loading stage.

**To Reproduce**
1. Run `npm start`
2. Load `localhost:3000/dash/admin` or `localhost:3000/dash/profile`. You will need to be logged in beforehand.
3. Note perpetual spinner
4. Click "Admin" or "Profile" in the header. Note the page loads properly now.
5. (optional) Make changes in a source file and save.
6. (optional) Go back to browser and note perpetual spinner again.

**Expected behavior**
`UserList` or `ProfileForm` respectively should be displayed instead of a spinner.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Chromium 83 and Firefox 77

**Additional context**
- It's not actually stuck in trying to load data --- no errors are being thrown, and if you inspect the HTML you'll see that the appropriate components are already rendered. However, `PrivateRoute` is stuck thinking it's in a loading state.
- This seems to be limited to routes with components that directly depend on the `useAuth()` hook. Calling `useAuth()` through another hook, e.g. `useAPILazyUnsafe()`, does not appear to trigger the issue.
- I've tracked the bug to 9a15855, with 5f90d0a being the last good revision. Will post more context as I debug.

Marking as P1 as this affects development.

tyxchen:
Fixes #64.

Also introduces a small improvement in how api handles uncontained server errors (which would be delivered in HTML) that I discovered in testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "jest": {
-    "coverageReporters": ["json", "text"]
+    "coverageReporters": [
+      "json",
+      "text"
+    ]
   },
   "dependencies": {
     "@babel/cli": "^7.10.1",
@@ -86,6 +89,7 @@
     "@types/express": "^4.17.6",
     "@types/http-proxy-middleware": "^0.19.3",
     "@types/morgan": "^1.9.0",
+    "@types/scheduler": "^0.16.1",
     "@types/serialize-javascript": "^1.5.0",
     "core-js": "^3.6.5",
     "eslint": "^6.8.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -96,6 +100,7 @@
     "prettier": "^2.0.5",
     "react-test-renderer": "^16.12.0",
     "rimraf": "^3.0.2",
+    "scheduler": "^0.18.0",
```
ktrieu, line 103: What's `scheduler` for?

tyxchen: It's used in conjunction with `React.Profiler` to trace updates. See: https://gist.github.com/bvaughn/8de925562903afd2e7a12554adcdda16

PR #68, opened 2020-07-17
New editor formatting

### LINKED ISSUE #39: Add basic paragraph and character styles to the editor
We currently have the bare minimum bold, italic, underline, and strikethrough, plus a link widget that I'm not completely happy with.

In InDesign right now, we have various levels of emphasis that automatically apply bold and italic. Do we want to abolish bold/italic etc. in favor of just these?

Disregarding that issue, the basics I'm shooting for:
- headers
- link formatting (both the styles and auto formatting text that matches)
- bold/italic, in some manner as discussed above
- strikethrough
- article references

ktrieu:
This PR cleans up the editor code, and adds headers, article references, and all six of our stress/emphasis styles.

Also undo/redo, thanks to `slate-history`.

Closes #39.


src/editor/components/styles/controls.scss
```diff
@@ -31,9 +31,28 @@
...
+    }
+  }
+}
+
 .control-popover {
   padding: $spacer;
   background: white;
-  border: 1px solid black;
+  border: solid 1px gray;
   border-radius: 5px;
```
tyxchen, line 56: Should use one of our standardized units

src/editor/__tests__/emph-stress.unit.test.ts
```diff
@@ -0,0 +1,317 @@
+import createCustomEditor from "../CustomEditor";
+import { Mark, BlockElementType } from "../types";
+import { increaseEmph, increaseStress } from "../helpers";
+
+describe("increaseEmph", () => {
```
tyxchen, line 5: Should test the invariant that only one emphasis or stress level can be active at any time. Ditto with increaseStress.

src/editor/helpers.ts
```diff
@@ -1,71 +1,226 @@
...
+  }
+};
+
+/**
+ * Returns true if `iterable` has no elements.
+ * @param iterable The iterable to check
+ */
+const isIterableEmpty = (iterable: Iterable<any>) => {
+  // this is weird but its faster than converting the iterable to an array then
+  // checking its length
```
tyxchen, line 79: My guess here would be that the JIT can unroll the loop, but it cannot unroll array creation. Also, array creation is a heap process, while iteration is on the stack. Hence the performance discrepancy.

src/editor/helpers.ts
```diff
@@ -1,71 +1,226 @@
...
+  } else {
+    editor.addMark(Mark.Emph1, true);
+  }
+};
+
+/**
+ * Returns true if `iterable` has no elements.
+ * @param iterable The iterable to check
+ */
+const isIterableEmpty = (iterable: Iterable<any>) => {
```
tyxchen, line 77: Suggestion: Add a check if the iterable is an array so you can do a simple length comparison, something like

```suggestion
const isIterableEmpty = (iterable: Iterable<any>) => {
  if (Array.isArray(iterable)) return iterable.length == 0; // or `if ("length" in iterable)`, though that may be slower
```

src/editor/helpers.ts
```diff
@@ -1,71 +1,226 @@
...
+  }
+  return true;
+};
+
+/**
+ * Returns true if the passed Location contains inlines, and false otherwise
+ * @param at The location to search at. If at is undefined, the current selection
+ * will be used instead.
+ */
+export const hasInlines = (editor: Editor, at?: Location) => {
```
tyxchen, line 92: editor parameter should be described in documentation

src/editor/helpers.ts
tyxchen: Actually ignore this as this will def be slower than loop unrolling.

src/editor/SluglineEditor.tsx
```diff
@@ -54,15 +58,25 @@ interface SluglineEditorProps {
   saveArticleContent: (content: Node[]) => void;
 }
 
+const SPELLCHECK_ENABLE_WAIT_MS = 500;
```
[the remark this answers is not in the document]
ktrieu: I was going to explain how the spellchecks are hiding the underlines, but I've since realized two things:
1. This solution doesn't work that well since changing spellcheck does not change it for text that isn't being completely re-rendered.
2. As someone who has gone far sighted far before their natural time, I maybe should not take my experience as universal.

I'll just remove it, I thnk, unless you have a more creative idea.

src/editor/helpers.ts
ktrieu: I was thinking weird as in "this is a weird piece of code to write, but here's why", but you're probably right.

src/editor/SluglineEditor.tsx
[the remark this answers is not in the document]
ktrieu: Removed: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/3a9a04bb25b325edab7c22863b402e93cb3aaf33

src/editor/SluglineEditor.tsx
```diff
@@ -107,6 +131,7 @@ const SluglineEditor: React.FC<SluglineEditorProps> = (
             placeholder="Start your masterpiece..."
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/babf6b1c339fd3a9f1672ae425cd0d4a5c4bb405

src/editor/components/Leaf.tsx
```diff
@@ -4,20 +4,56 @@ import { Mark } from "../types";
...
-  }
-  if (props.leaf[Mark.Italic]) {
+  if (props.leaf[Mark.Emph1]) {
     children = <i>{children}</i>;
   }
-  if (props.leaf[Mark.Underline]) {
-    children = <u>{children}</u>;
+  if (props.leaf[Mark.Emph2]) {
+    children = (
+      <b>
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/7241af6b2b214c9e859b24bd30940983cf5e51c6

src/editor/components/Link.tsx
```diff
@@ -3,7 +3,19 @@ import { RenderElementProps, useEditor, ReactEditor } from "slate-react";
 import { LinkElement } from "../types";
 import LinkPopover from "./LinkPopover";
 import { Transforms } from "slate";
-import PopoverWrapper from "../../shared/PopoverWrapper";
+import PopoverWrapper from "./PopoverWrapper";
+
+export const formatHref = (href: string) => {
+  if (href.includes("@")) {
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/dfb7bdf5138cae5c922c6e1b4b074a9863235e6c

src/editor/components/Link.tsx
```diff
@@ -3,7 +3,19 @@ import { RenderElementProps, useEditor, ReactEditor } from "slate-react";
...
+    if (!href.startsWith("mailto:")) {
+      return "mailto:" + href;
+    }
+  } else if (!href.startsWith("http://") && !href.startsWith("https://")) {
+    return "https://" + href;
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/ccd502ddee386a46beef2e2bdb2e86e47e693cf6

src/editor/components/Link.tsx
```diff
@@ -35,7 +47,7 @@ const Link: React.FC<RenderElementProps> = (props: RenderElementProps) => {
         onClick={() => {
           setShow(true);
         }}
-        href={element.href}
+        href={formatHref(element.href)}
```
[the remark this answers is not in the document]
ktrieu: You have a good point. The footnote solution seems like the best way, though we'd then have to rearrange people's user-created footnotes on export.

src/editor/helpers.ts
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/fc1654257f05447b3832c763b6026083e3ba690e

src/editor/__tests__/emph-stress.unit.test.ts
ktrieu: I've replaced this with a general "all marks are mutually exclusive" feature in the editor: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/fc1654257f05447b3832c763b6026083e3ba690e

src/editor/components/styles/controls.scss
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/2a2cfeeea2d65a5ecf4a4b0c011de9c1175138c9

src/editor/components/controls.tsx
```diff
@@ -1,14 +1,72 @@
...
+  );
+};
+
+export const IncreaseStressButton: React.FC = () => {
+  const editor = useSlate();
+  return (
+    <button
+      className="editor-control"
+      onClick={() => {
+        increaseStress(editor);
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/484dd6213ba5be3968d104a4fd859b423acc4e89

src/editor/components/controls.tsx
```diff
@@ -1,14 +1,72 @@
...
+  );
+};
+
+export const IncreaseEmphButton: React.FC = () => {
+  const editor = useSlate();
+  return (
+    <button
+      className="editor-control"
+      onClick={() => {
+        increaseEmph(editor);
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/484dd6213ba5be3968d104a4fd859b423acc4e89

ktrieu:
I suspect the weird issues around applying marks was an artifact of us not focusing the editor after applying them, and that fix should have resolved them. I've been unable to reproduce them after the fix, but if you can still repro, let me know.

src/editor/components/controls.tsx
```diff
@@ -20,11 +69,12 @@ export const ToggleMarkButton: React.FC<ToggleMarkButtonProps> = (
 ) => {
   const editor = useSlate();
   const active = isMarkActive(editor, props.mark);
-  const className = active ? "editor-control active" : "editor-control";
+  const className = active ? `editor-control active` : "editor-control";
```
tyxchen, line 72: Would be nice if there could be some indeterminate state for the dropdown buttons that is activated when a dropdown item is active (so for example, the emphasis button would show this indeterminate state if emphasis 4 is selected).

ktrieu:
> Something else I'd like to add: I think the input for author name should look like an input, as right now it looks like a paragraph. I think this is something we can save for the eventual redesign though.

What do you think about moving it to the right, which matches how we do it in print? I've attached a screenshot:

image

src/editor/components/Link.tsx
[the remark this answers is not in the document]
tyxchen: I think the rearrangement is possible since we're working with a JSON model after all.

tyxchen:
It's better, but it still doesn't solve the core UX problem that the input doesn't look like an input IMO. I think it's fine for the rest of the editor since there are very distinct placeholder states that cue the user towards the fields being editable, but when you start a new article the author is already prefilled with your writer name. 

An option is to have a sort of "sticky" author name field. The author name is pre-filled and pre-formatted, and always appears at the end of the article. Something like this:

image

And when you add content, it scrolls down like so:

image

Sidenote, I discovered that if you add content to a new article and then clear the buffer, the placeholder will not show anymore.

ktrieu:
So Slate supports editor normalization, where we can enforce any constraint on the document we want. This could let us declare our own "author" block that always has to be present and always has to be at the end of the document. There's an example [on their website](https://www.slatejs.org/examples/forced-layout).

This also lets us render footnotes after the author, as our lord intended.

That being said, I think we can save this for after this already considerable PR, and file an issue to handle it later.

On the placeholder thing, I'm selecting all and deleting, and manually backspacing away the text, and the placeholder still shows up.  Might need more detailed instructions to repro. I remember running into an issue like this a while ago, and from what I can recall, it's about nodes not getting properly removed and tricking the editor into believing there's still content.

tyxchen:
For the placeholder, try putting in nothing but line breaks, and then try clearing the textbox. That seems to be one way to repro.

src/editor/EditorControls.tsx
```diff
@@ -3,22 +3,40 @@ import {
...
+        <ToggleMarkButtonText text="Emphasis 3" mark={Mark.Emph3} />
+        <ToggleMarkButtonText text="Emphasis 4" mark={Mark.Emph4} />
+      </EditorDropdown>
       <ToggleMarkButton icon="strikethrough" mark={Mark.Strikethrough} />
       <ToggleMarkButton icon="code" mark={Mark.Code} />
       <div className="editor-separator" />
       <LinkButton />
       <InlineLatexButton />
+      <div className="editor-separator" />
+      <ToggleBlockButton icon="heading" blockType={BlockElementType.Header} />
```
tyxchen, line 38: This should be a dropdown for both heading styles.

ktrieu:
> Sorry to add on to the PR, but it just occurred to me that the icons in the dropdowns vs normal buttons are of different sizes. Also, see attached comment on heading styles.

Maybe I'm misunderstanding, but the icons seem to be the same size in the dropdowns and the regular buttons on my screen.

src/editor/EditorControls.tsx
```diff
@@ -3,22 +3,28 @@ import {
...
 const EditorControls: React.FC = () => {
   return (
     <div className="editor-controls">
-      <ToggleMarkButton icon="bold" mark={Mark.Bold} />
-      <ToggleMarkButton icon="italic" mark={Mark.Italic} />
-      <ToggleMarkButton icon="underline" mark={Mark.Underline} />
+      <IncreaseStressButton />
+      <IncreaseEmphButton />
       <ToggleMarkButton icon="strikethrough" mark={Mark.Strikethrough} />
       <ToggleMarkButton icon="code" mark={Mark.Code} />
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/9d47c8cfab90fdeb430c549eb2beb72165248aed
to
https://github.com/UWmathNEWS/slugline-web/pull/68/commits/f39a49307a6d2908ab6536b9a8d7f0e9122dedcb

covers this.

src/editor/components/controls.tsx
```diff
@@ -1,14 +1,72 @@
...
+    >
+      <ToggleMarkButtonText mark={Mark.ArticleRef} text="Article Reference" />
+    </DropdownButton>
+  );
+};
+
+export const IncreaseStressButton: React.FC = () => {
+  const editor = useSlate();
+  return (
+    <button
```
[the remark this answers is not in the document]
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/64820ec0fce34393bf07b6c560240023fd5dd23a

src/editor/EditorControls.tsx
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/b664fba667ae40e03a7009134ff8b8ce381f3ffa

ktrieu:
> For the placeholder, try putting in nothing but line breaks, and then try clearing the textbox. That seems to be one way to repro.

After looking into it, this appears to be a known Slate issue: see https://github.com/ianstormtaylor/slate/issues/3432 and https://github.com/ianstormtaylor/slate/issues/3605.

I'll keep watch on those and see if its resolved. In the worst case, we can disable the default placeholder and build our own, which shouldn't be too onerous.

src/editor/components/controls.tsx
ktrieu: I've done this for the `EditorDropdown` here: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/1cb67d5d34e0d49fe4ca701ac1151b41aac8b88e, but I'm not quite sure how this would work for the 
"More" dropdown considering its actually one button and not two like the others are. 

Just inverting the colors like we do for the formatting buttons looks kind of ugly in my opinion.

tyxchen:
> > Sorry to add on to the PR, but it just occurred to me that the icons in the dropdowns vs normal buttons are of different sizes. Also, see attached comment on heading styles.
> 
> Maybe I'm misunderstanding, but the icons seem to be the same size in the dropdowns and the regular buttons on my screen.

Not sure if this is due to relative icon sizing, but the bold and italic icons are much larger than the strikethrough icon:

image

tyxchen (changes requested):
Sorry for the tardy review. Things have loosened up again so I can recommit time into this.

Just a couple more issues:

1. When the cursor is at the end of a heading block, pressing enter should create a new paragraph, and should not continue the heading style.
2. Undoing and redoing seems to be really iffy when undoing/redoing operations on a selection. So far, the most consistent way I've been able to reproduce this is by typing out a sentence of text, making selections in the middle of words and formatting them, and after the third select-format loop, undo the operations and watch the selections go wack.
3. In EditorDropdown, should the main button be active when a substyle is active, and should clicking on the main button deactivate the style? Right now it does feel clunky going into the dropdown and deactivating the active style, though this may be a decision that should wait for user feedback.

ktrieu:
> > > Sorry to add on to the PR, but it just occurred to me that the icons in the dropdowns vs normal buttons are of different sizes. Also, see attached comment on heading styles.
> > 
> > 
> > Maybe I'm misunderstanding, but the icons seem to be the same size in the dropdowns and the regular buttons on my screen.
> 
> Not sure if this is due to relative icon sizing, but the bold and italic icons are much larger than the strikethrough icon:
> 
> image

Something fishy is going on, because it looks like this for me: 
image

tyxchen:
> > > > Sorry to add on to the PR, but it just occurred to me that the icons in the dropdowns vs normal buttons are of different sizes. Also, see attached comment on heading styles.
> > > 
> > > 
> > > Maybe I'm misunderstanding, but the icons seem to be the same size in the dropdowns and the regular buttons on my screen.
> > 
> > 
> > Not sure if this is due to relative icon sizing, but the bold and italic icons are much larger than the strikethrough icon:
> > image
> 
> Something fishy is going on, because it looks like this for me:
> image

Alright, so the discrepancy appears to be a Firefox thing, and furthermore, is linked to the amount of horizontal padding associated to buttons with Firefox on Linux. Specifically, Chrome gives buttons 6px of horizontal padding, while Firefox gives 8px. Since we restrict the width of buttons in the editor, this causes wide icons to shrink in Firefox, as then the dominating size restriction becomes the available width.

One fix is to remove the horizontal padding on .editor-control so that icon sizes are universally restricted by height. This might cause the icons to be a bit larger than we may be comfortable with, but that can be easily changed.

ktrieu:
> When the cursor is at the end of a heading block, pressing enter should create a new paragraph, and should not continue the heading style.

I've added code to reset the paragraph type whenever a new paragraph is created: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/276d10f9918ef4fff4dd8897f1f0213b2c6525f9

> button shenanigans

For this, I've standardized the padding to `6px`, which is the Chrome value: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/dde316541423af679a612bdfd54cb08566a48947

I tried the "no horizontal padding" method, but that made wide icons like the code block get uncomfortably close to the edges of the button.

ktrieu:
In terms of the outstanding weird editor behaviours, I've created two tracking issues for them: https://github.com/UWmathNEWS/slugline-web/issues/76 and https://github.com/UWmathNEWS/slugline-web/issues/77. Since they're mostly Slate bugs anyway, I'd recommend that we save any fixes we have for those for another PR.

src/editor/components/styles/controls.scss
```diff
@@ -8,6 +8,7 @@
   background: none;
   border-radius: $border-radius;
   margin: spacing(1) / 2;
+  padding: 6px;
```
tyxchen, line 11: Might just be good to leave a comment here for why we've set it to 6px specifically.

ktrieu: Until we pass a UN resolution making Chrome the only legal browser for consumer use, I've added a comment explaining things: https://github.com/UWmathNEWS/slugline-web/pull/68/commits/2af1588adeba9c877b4d77a68183da0275b26c85

tyxchen (approved):
lgtm :+1:

PR #73, opened 2020-07-30
More code scanning

vincentmacri:
Adds some code scanning tool GitHub recommended. Might be useful, let's see what it does.

### DIFF
--- a/.github/workflows/ossar-analysis.yml
+++ b/.github/workflows/ossar-analysis.yml
@@ -1,0 +1,45 @@
+# This workflow integrates a collection of open source static analysis tools
+# with GitHub code scanning. For documentation, or to provide feedback, visit
+# https://github.com/github/ossar-action
+name: OSSAR
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  OSSAR-Scan:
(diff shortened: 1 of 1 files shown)


.github/workflows/ossar-analysis.yml
```diff
@@ -0,0 +1,45 @@
+# This workflow integrates a collection of open source static analysis tools
+# with GitHub code scanning. For documentation, or to provide feedback, visit
+# https://github.com/github/ossar-action
+name: OSSAR
+
+on:
+  push:
+  pull_request:
```
tyxchen, line 8: Should these keys be set?

vincentmacri: This is the default workflow for the tool. I believe this sets it to run on all pushes and PRs.

tyxchen: Wonder if we should have it for pushes and PRs to master only, as we do for the other security workflows?

vincentmacri: I'd leave it on the default settings first and change later if needed.

tyxchen (approved):

PR #78, opened 2020-09-15
Implement home

### LINKED ISSUE #57: Implement home page
See the Figma for mockups (can also be seen in Wiki > Design).

## Ideas

- Have the lead take on the most recent issue's colours
- Randomized taglines:
    ```
    Fresh off the press
    Recently proved to exist
    Newly conjectured
    Hotter than hot pink
    Hot and ready
    Straight from the pizza oven
    ```
- Extract cover page of PDFs as preview image (needs to be backend)
- Infinite loading at bottom or link to all issues?

tyxchen:
-- Fixes #57 --

Implements the home page. Sibling to UWmathNEWS/slugline-api#26.


ktrieu:
Didn't add my tagline from Discord. Rejected.

ktrieu:
> Didn't add my tagline from Discord. Rejected.

Ok, I missed it in the list. Indignant rage retracted.

tyxchen:
> > Didn't add my tagline from Discord. Rejected.
> 
> Ok, I missed it in the list. Indignant rage retracted.

I even credited you and everything :(

src/shared/components/Button.tsx
```diff
@@ -0,0 +1,50 @@
...
+    </BsButton>
+  );
+});
+
+export const LinkButton = React.forwardRef<
+  HTMLAnchorElement,
+  ButtonProps & LinkProps
+>(({ variant, className, children, ...props }, ref) => {
+  return (
+    <Link
```
ktrieu, line 40: Would it be possible to use the `as` prop on the react-bootstrap button for this use case?

src/shared/components/Paginator.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <li className={`Paginator_item ${className || ""}`}>
+      <Link {...props} className="Paginator_link">
+        {children}
+      </Link>
+    </li>
+  );
+};
+
+class Paginator<T> extends React.Component<
```
ktrieu, line 23: Curious as to why this is a class component?

src/shared/components/Paginator.tsx
```diff
@@ -0,0 +1,106 @@
...
+              to={url!(pagination.page - 1)}
+              className="Paginator_prev"
+              title="Previous page"
+              aria-label="Previous page"
+            >
+              <FontAwesomeIcon icon="chevron-left" />
+            </PaginatorItem>
+          )}
+          {new Array(pagination.num_pages).fill(0).map((_, page) => {
+            if (
```
ktrieu, line 48: A comment summarizing this large if condition would be helpful.

src/shared/components/Paginator.tsx
```diff
@@ -0,0 +1,106 @@
...
+            <PaginatorItem
+              to={url!(pagination.page - 1)}
+              className="Paginator_prev"
+              title="Previous page"
+              aria-label="Previous page"
+            >
+              <FontAwesomeIcon icon="chevron-left" />
+            </PaginatorItem>
+          )}
+          {new Array(pagination.num_pages).fill(0).map((_, page) => {
```
ktrieu, line 47: This whole map function could be simplified by renaming `page` to `idx` and using `page` = `idx + 1`.

src/home/Home.tsx
```diff
@@ -1,14 +1,309 @@
...
+}) => {
+  return (
+    <div className={`IssueEntry ${className || ""}`}>
+      <IssueEntryBody
+        issue={issue}
+        dateline={`Volume ${issue.volume_num} Issue ${issue.issue_code}`}
+      />
+      <div className="mt-3">
+        {/* TODO: When issue interface is changed, remove fallback */}
+        <ActionLink to={issue.pdf || ""} className="IssueEntry_cta">
```
ktrieu, line 98: The `FullHero` links to the issue page, but the link for these regular entries links direct to the pdf. Is this intentional?

ktrieu: Also worth noting since this renders to a `<Link>` component, the `issue.pdf` gets treated as a relative URL, and we currently return them as absolute URLs, which breaks the link.

src/shared/helpers.tsx
```diff
@@ -123,3 +123,20 @@ export const renderRoutes: {
...
+ */
+export const coverSrc = (issue: Issue, size: number, type?: "RGB" | "LA") =>
+  issue.pdf +
+  `.COVER-${type ?? (issue.colour === "paper" ? "RGB" : "LA")}-${size}x.png`;
+
+/**
+ * Does nothing.
+ */
+/* istanbul ignore next */
+export const noop = () => {};
```
ktrieu, line 142: Can't find any references to this function. What is it for?

ktrieu (changes requested):
This looks really good. Like, it matches the vision I had in my head when I started thinking about this around 300 years ago.

It's really late now, so I will get to the API PR tomorrow.

src/home/Home.tsx
tyxchen: The link discrepancy is an oversight on my part, my bad.

Regarding `issue.pdf`, I looked into it and the API returns a relative link for the `latest` endpoint, but an absolute link on the list endpoint. Would you happen to know why?

src/home/Home.tsx
```diff
@@ -1,14 +1,309 @@
...
+  return (
+    <>
+      <Dateline>
+        {dateline} &bull;{" "}
+        {format(new Date(issue.publish_date || Date.now()), "d MMM y")}
+      </Dateline>
+      <h2 className="IssueEntry_title mt-1">
+        <Link to={`/issues/${issue.id}`}>{issue.title}</Link>
+      </h2>
+      <div className="IssueEntry_description mt-3">{issue.description}</div>
```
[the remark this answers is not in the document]
tyxchen: Yes, the class names are all for futureproofing.

src/home/Home.tsx
ktrieu: Seems to be [this](https://stackoverflow.com/a/52357351). The `ViewSet`s pass in a context with the Django request automatically, which lets DRF build an absolute URL. The `latest` endpoint uses a manually instantiated serializer, so it can't do this. 

There's an easy fix in the answer, basically we pass in `{ 'request': request }` to the serializer on instantiation.

src/home/Home.tsx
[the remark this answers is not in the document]
ktrieu: I don't know if I'm a fan of that, I think they imply custom styling when there currently isn't any. It seems relatively simple to add back a class name if they become needed.

src/shared/components/Button.tsx
tyxchen: I've tried, but it breaks the typings on refs.

src/shared/components/Paginator.tsx
tyxchen: Added documentation: 9a1440ba

src/shared/components/Paginator.tsx
tyxchen: Added: 9a1440ba

src/shared/components/Paginator.tsx
tyxchen: Done: 9a1440ba

src/home/Home.tsx
tyxchen: Fixed the links: ca7e995b

src/shared/helpers.tsx
tyxchen: It's a no-op function, we have a bunch of empty functions that are causing test coverage to fail because we can't actually test them (e.g. in Auth.ts)

src/shared/components/Button.tsx
```diff
@@ -0,0 +1,50 @@
...
+  variant:
+    | "dark"
+    | "light"
+    | "pink-dark"
+    | "pink-light"
+    | "grey-dark"
+    | "grey-light";
+}
+
+const Button = React.forwardRef<
```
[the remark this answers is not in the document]
tyxchen: Eventually, yes.

src/home/Home.tsx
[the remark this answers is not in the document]
tyxchen: Personally I'd rather have classes even if they're unused for styling. I find the semantic meaning they give helpful, and it also helps with BEM.

src/home/Home.tsx
```diff
@@ -1,14 +1,309 @@
...
+    </>
+  );
+};
+
+const IssueEntry: React.FC<{ issue: Issue } & ForwardAttributes> = ({
```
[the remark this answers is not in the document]
tyxchen: Refactored: f45bffbd

src/shared/components/Paginator.tsx
tyxchen: Made moot with 38007020

ktrieu (approved):

ktrieu:
Alright, looks good to go. Appreciate the docs.

PR #80, opened 2020-09-18
Update react-bootstrap to 1.3.0 and typescript to 3.9

tyxchen:
See title.

This also means we can start using `import type` declarations and top-level await, the former which might also reduce our bundle size slightly.

### DIFF
--- a/src/shared/form/util.ts
+++ b/src/shared/form/util.ts
@@ -1,5 +1,5 @@
-import { FormContextValues } from "react-hook-form";
-import { APIError } from "../types";
+import type { FormContextValues } from "react-hook-form";
+import type { APIError } from "../types";
 
 export const cleanFormData = <T extends {}>(data: T): Partial<T> => {
   let ret: Partial<T> = {};
--- a/src/shared/form/util.ts
+++ b/src/shared/form/util.ts
@@ -1,10 +1,14 @@
 import { FormContextValues } from "react-hook-form";
 import { APIError } from "../types";
 
-export const cleanFormData = <T>(data: T): Partial<T> => {
-  return Object.fromEntries(
-    Object.entries(data).filter(([key, value]) => value !== "")
-  );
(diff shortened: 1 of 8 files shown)


ktrieu:
I've been unable to get this to compile on Windows or Mac. An `npm install` followed by an `npm run start` gives me the following error:

```
Failed to compile.

./src/routes/Public.tsx
  Line 1:12:  Parsing error: '=' expected
```

which makes me think that Typescript is choking on the new import type syntax. Curiously, I seem to have `typescript@3.9.7` installed. Is there anything special you did to get it work on your side?

tyxchen:
Strange, I'm also on 3.9.7. Can you check your react-scripts version? In the meantime I'll see if fixing it at 3.8.3 changes anything.

tyxchen:
Seeing the install issue was fixed, I'm going to go ahead and move to TS 3.9 (since it seems we support it anyways and it gives a lot of cool stuff)

src/shared/form/util.ts
```diff
@@ -1,10 +1,14 @@
-import { FormContextValues } from "react-hook-form";
-import { APIError } from "../types";
+import type { FormContextValues } from "react-hook-form";
+import type { APIError } from "../types";
 
-export const cleanFormData = <T>(data: T): Partial<T> => {
-  return Object.fromEntries(
-    Object.entries(data).filter(([key, value]) => value !== "")
-  );
+export const cleanFormData = <T extends {}>(data: T): Partial<T> => {
```
ktrieu, line 4: Can you clarify what the differences are between this function and the old one?

tyxchen: Functionally, there's no difference. The reason for the change is that Typescript changed the typing of `Object.entries` to return the equivalent of `{ [k: any]: any }` instead of `{ [k keyof T]: T[k] }`. Now the only way to iterate over an object and ensure keys are typed `keyof T` is with a for-in loop, hence the change.

ktrieu (approved):

PR #82, opened 2020-09-21
Add list support

ktrieu:
This PR adds ordered and un-ordered lists, which enables the all-important N things listicle.

Besides the various boilerplate types a new block type needs, it also adds some additional logic to account for the fact that lists require a level of nesting.

Screen Shot 2020-09-20 at 9 13 17 PM


tyxchen (approved):
I have been convinced by your image.

ktrieu:
I knew it would work.

PR #84, opened 2020-11-07
Add image (and void block support) to the editor

ktrieu:
This PR adds support for images, and void blocks more generally, to the editor. 

This ended up being a little more complicated than I thought because of the difficulties of inserting paragraphs before/after void blocks. To solve this, we use void spacers, which are void blocks that we automatically insert before and after void content blocks. These provide "anchors" for the selection to grab and insert paragraphs in.

I've said this before, but this is one of the last architectural issues to solve in the editor. We have now covered all four permutations of the block/inline void/not void universe. Addition of new, fancy block types should be much simpler now.

The image block here just take a link as a property and show an `` element. If the image doesn't load, we show a nice error block instead of a broken tag.


ktrieu:
Turns out the security checks are failing because of some `react-scripts` dependencies. I'd update to 4.0.0, but there's a [fatal bug](https://github.com/facebook/create-react-app/issues/9868) with CRA and Typescript, so I'll wait for the fix for that to drop. In the meantime, feel free to review.

tyxchen:
Oh no, CRA.....

Let's leave the CRA upgrade for a separate PR. We can bundle upgrading to Typescript 4.x with that, which gives us some really nice language features to boot.

src/editor/CustomEditor.ts
```diff
@@ -55,6 +52,10 @@ const createCustomEditor = () => {
     switch (e.type) {
       case InlineElementType.InlineLatex:
         return true;
+      case BlockElementType.Image:
+        return true;
+      case BlockElementType.VoidSpacer:
+        return true;
```
tyxchen, line 58: This condition would probably be better expressed as returning a boolean expression.

src/editor/CustomEditor.ts
```diff
@@ -98,10 +101,102 @@ const createCustomEditor = () => {
...
+    if (Editor.string(editor, editor.selection.anchor.path) !== "") {
+      return false;
+    }
+    return true;
+  };
+
+  const deleteForwardCustom: typeof deleteForward = (unit) => {
+    if (
+      editor.selection &&
+      Range.isCollapsed(editor.selection) &&
```
tyxchen, line 128: This particular `editor.selection && Range.isCollapsed(editor.selection)` pattern appears a few times, do we want to wrap this into a function?

src/editor/components/styles/Image.scss
```diff
@@ -0,0 +1,37 @@
...
+    margin-left: auto;
+    margin-right: auto;
+
+    max-width: 85%;
+    min-width: 50%;
+
+    height: auto;
+
+    &--selected {
+      box-shadow: 0px 0px 0px 2px lightblue;
```
tyxchen, line 19: Is this a common selection style? If so, would this be better expressed as a mixin or pseudo-selector?

src/editor/helpers.ts
```diff
@@ -170,42 +172,77 @@ export const isBlockActive = (editor: Editor, blockType: BlockElementType) => {
...
-      { type: blockType },
-      { mode: "all", match: (node) => Editor.isBlock(editor, node) }
-    );
-  }
+    // if the requested block is a list and that list is currently not active,
+    // we are toggling a list ON, and we have to do some wrapping
+    if (isListType(blockType) && !isActive) {
+      wrapListItems(editor, blockType);
+    }
+    // otherwise, toggle as usual
```
tyxchen, line 197: nit: probably should move the comment within the else-if block for style consistency

src/editor/normalize.ts
```diff
@@ -0,0 +1,135 @@
...
+      editor,
+      {
+        type: BlockElementType.VoidSpacer,
+        children: [{ text: "" }],
+      },
+      {
+        at: [editor.children.length],
+      }
+    );
+    return;
```
tyxchen, line 89: Are the returns here and one line 101 necessary?

src/editor/normalize.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+};
+
+/**
+ * Normalizes the editor, i.e., the top-level container that holds all blocks.
+ * This handles top-level layout.
+ * @param editor The editor to normalize
+ */
+export const normalizeEditor = (editor: Editor) => {
+  for (let idx = 0; idx < editor.children.length; idx++) {
```
tyxchen, line 112: Suggestion: use Array.prototype.entries() for iteration, e.g.

```js
for (const [idx, node] of editor.children.entries()) {
  ...
```

src/editor/normalize.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export const normalizeBlock = (
+  editor: Editor,
+  block: BlockElement,
+  path: Path
+) => {
+  // remove lists with no children
+  if (isListType(block.type) && block.children.length === 0) {
+    Transforms.removeNodes(editor, { at: path });
+    return;
```
tyxchen, line 133: This return seems extraneous.

tyxchen (commented):
Here's a first round of comments, sorry for the long delay. I still haven't pulled this onto my local to test, unfortunately I'm not sure when's the soonest I can get to that.

tyxchen:
Just pulled and tested locally. Behaviour looks good to me, just a few thoughts:

- We may want to pull up an error or warning if the image is non-https (rare nowadays, but still possible)
- If possible, we may want to warn of images from sources that blacklist us (e.g. Discord) as well as images with a local URI (e.g. file://, blob: or localhost, because who knows?)

src/editor/normalize.ts
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/47f43914bb06fc8a929d8d02739d83683829f688

src/editor/normalize.ts
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/47f43914bb06fc8a929d8d02739d83683829f688

src/editor/normalize.ts
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/47f43914bb06fc8a929d8d02739d83683829f688

src/editor/helpers.ts
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/47f43914bb06fc8a929d8d02739d83683829f688

src/editor/normalize.ts
ktrieu: This actually breaks tests, oops.

Because we're actually editing the list of children as we iterate through them, `.entries()` won't work, since it gathers the index and children in one shot.

I can add a comment to this effect if you'd like.

Side note: There's a new feature in a later version of Slate that should remove the need for this function entirely, but I'll save that for another PR.

src/editor/CustomEditor.ts
ktrieu: Good catch: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/b94d41ba74a36664f8296c6892f5282b6108ccac

src/editor/CustomEditor.ts
ktrieu: I wanted to leave it open for adding more void blocks in the future, in which case a boolean would get unwieldy.

src/editor/components/styles/Image.scss
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/a257e0d1fafbea9d55ec1c8df5f29c3313bdbcca

ktrieu:
Fixed your comments, but we're waiting on https://github.com/ianstormtaylor/slate/pull/4050 upstream to fix that last security issue.

src/editor/normalize.ts
```diff
@@ -98,13 +97,12 @@ const normalizeVoidBlock = (editor: Editor, path: Path) => {
...
         { at: Editor.before(editor, nextPath), mode: "highest" }
       );
-      return;
     }
   }
 };
 
 /**
- * Normalizes the editor, i.e., the top-level container that holds all blocks.
+ * Normalizes the editor, i.e.,q the top-level container that holds all blocks.
```
tyxchen, line 105: Q

src/editor/normalize.ts
tyxchen: Ahh, I didn't catch that, my bad. (sidenote: imagine if Typescript had const assertions for arguments...the joy....)

src/editor/CustomEditor.ts
tyxchen: Would it be? I was under the impression that Prettier would make long expressions multiline, which would help with the readability.

ktrieu: Maybe readability is the wrong word, I think it would be clearer to use a switch on block types than parsing a long multi-line conditional.

I like the fall-through idea, I'll implement that.

src/editor/normalize.ts
ktrieu: I swear I removed this. https://github.com/UWmathNEWS/slugline-web/pull/84/commits/a76cd9f0a23289f54e872d958c0e80885957c72f

src/editor/CustomEditor.ts
ktrieu: https://github.com/UWmathNEWS/slugline-web/pull/84/commits/fa74ea724f50b9f423107f681f1cb70d5c498384

ktrieu:
I was thinking about the extra image URL checks you mentioned, and I realized that if we stuck with our "upload to Imgur" policy for images, we'd basically be hotlinking en masse when the website goes public, which is against Imgur TOS, and probably a lot of other places.

The answer here is probably "build an image uploading system to our servers". Thoughts?

tyxchen:
An image uploading system makes me skittish because file upload functionality is a common source of major security issues, in addition to opening up the possibility of abuse.

I wasn't aware that Imgur's TOS prohibited hotlinking. This will be a topic I need to put more thought into.

ktrieu:
The other option is some sort of service like S3, which would save us development time and alleviate those security issues, at the cost of having to manage the hosting service instead, plus some monthly cost, which will be annoying financially.

tyxchen:
Yeah, the cost of S3 increases over time (pricing is based on amount stored + number of accesses). I also don't know how payment would work out, and long-term there's a pretty high risk that this cost might be forgotten or overlooked by future editorial teams, which could lead to some nasty surprises.

tyxchen:
@ktrieu It looks like the slate security issue is fixed

ktrieu:
Looks like we have another, different security problem: https://github.com/facebook/create-react-app/issues/10945

CRA uses PostCSS 7, which has a security problem which PostCSS doesn't want to patch because they are now supporting PostCSS 8 only. I figure I had about a five-day window to merge this between the Slate thing being fixed and the CRA thing breaking.

tyxchen:
Ugh.

I'm beginning to wonder if we should drop CRA completely. It seems to just be giving us headache after headache, for really not all that much benefit.

tyxchen:
So it looks like a whole bunch of RegEx DoS issues got found/discovered recently. Ugh.

At this point I'm inclined to just merge and get this over with, thoughts?

ktrieu:
I think we can mitigate this kind of thing by moving `react-scripts` to `devDependencies` and switching the security check command to `npm audit --production`. Then, we can skip these warnings which don't really represent vulnerabilities in our application while hopefully still being informed of actual ones that may come up.

I tried that, but there are still some vulnerabilities in some other packages. Since this PR is so huge and it's been stuck for so long, let's just merge it and I'll make those changes and deal with those other vulnerabilities in another PR.

tyxchen (approved):

tyxchen:
Builds are failing because we moved `src/styles/slugline.scss` to `src/styles/primary_style.scss`, `src/styles/_slugline_vars.scss` to `src/styles/_primary_style_vars.scss`, etc. Probably want to update that before merge.

